PR #71, opened 2016-07-12
Python requirements file detection improvements

Kami:
I wanted to add support for requirements files in the following format: `*-requirements.txt`.

It turns out this format is already support since the regular expression match function checks for match anywhere in the string and doesn't require input string to start with `require` (aka `^require`).

If nothing else, the tests should at least still be useful :)


andrew:
Great, thanks I'll look at deploying that later today or tomorrow once the traffic has died down a little.

Kami:
Looks like the test failure is not related to my change, but to something which perhaps looks like deterministic test (it seems like that list needs to be sorted, but I didn't really dig in much) :)

andrew:
Oh yeah, very strange that the order of that list has changed, but won't break anything, I'll get that fixed up, thanks again

PR #72, opened 2016-07-12
teach the nuget parser how to handle paket lockfiles

baronfel:
[Paket](http://fsprojects.github.io/Paket/) is an alternative Nuget client for the .net ecosystem that also handle git and file-based dependencies.  This PR teaches the Nuget parser how to read the [lockfiles](http://fsprojects.github.io/Paket/lock-file.html) generated by paket, which contain every dependency across the entire solution.

I'm not sure how detailed the dependency descriptions are, but this just handles the most common use case for paket: Nuget dependencies.  If there is enough demand I could expand the parser to learn about git, github, file, and group dependencies.

### DIFF
--- a/lib/bibliothecary/parsers/nuget.rb
+++ b/lib/bibliothecary/parsers/nuget.rb
@@ -2,6 +2,7 @@
 # Project.json
 # Project.lock.json
 # *.nuspec
+# paket.lock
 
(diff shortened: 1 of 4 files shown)


andrew:
Brilliant, thanks! I'll get that deployed shortly

PR #101, opened 2016-11-11
Only report dependencies that have crates.io sources.

jdm:


jdm:
It's been 7 years since the last time I wrote any Ruby code, so I'm happy to fix anything unidiomatic.

andrew:
Great work, thanks, I'll get that deployed shortly

PR #405, opened 2017-12-05
Make test less fragile

katzj:
This test was depending on the state of the repo pretty heavily
and needed to be updated whenever deps were updated. There's already
a good framework for testing specific parsers against fixtures so
we don't need to do that in this test also.


katzj:
The swift parser test seems to be failing sporadically but that's with or without this

PR #407, opened 2017-12-20
CLI interface for listing dependencies in a given path

andrew:
Just one command for now, but using [commander](https://github.com/commander-rb/commander) so it's easy to add more later:

```
$ bibliothecary list --path path/to/my/repository
```
 
screen shot 2017-12-20 at 09 59 24


andrew:
🚢 in v5.6.0

PR #417, opened 2018-03-28
Enable mono repo support by tweaking regexes to allow sub directories

mellisio:

- [x] Is there a corresponding ticket for your pull request?
- [x] Have you written new tests for your changes?

### DIFF
--- a/lib/bibliothecary/analyser.rb
+++ b/lib/bibliothecary/analyser.rb
@@ -51,21 +51,17 @@ module Bibliothecary
       end
 
       def analyse_contents(filename, contents)
-        begin
-          dependencies = parse_file(filename, contents)
-          if dependencies && dependencies.any?
-            {
-              platform: platform_name,
-              path: filename,
-              dependencies: dependencies,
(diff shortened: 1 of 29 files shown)


andrew:
@mellisio Looking good, only thing that I'm cautious of is picking up things embedded inside of vendored node_modules folders but I have a feeling that those will be filtered out elsewhere, just the removal of the tests makes me doubt myself.

mellisio:
@andrew I was thinking, perhaps erroneously, that those folders shouldn't be included in git and wouldn't affect our use case since we do a git pull every time.

andrew:
@mellisio usually they wouldn't but some people do include them to avoid depending on npmjs.org in production, which then throws a huge amount of extra manifest files into the repo

andrew:
Checked this morning and `node_modules` and `bower_components` folders are ignored by the default configuration that I added a few months back.

Just released this as 6.3.0: https://rubygems.org/gems/bibliothecary/versions/6.3.0

PR #419, opened 2018-05-17
Raise on bad response from parse host rather than silently passing []

mellisio:

### DIFF
--- a/lib/bibliothecary/exceptions.rb
+++ b/lib/bibliothecary/exceptions.rb
@@ -1,0 +1,3 @@
+module Bibliothecary
+  class ParseHostError < StandardError; end
+end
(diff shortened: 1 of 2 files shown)


andrew:
We should separate out the difference between how we handle/report parsing errors because of a malformed or unknown format and other kinds of errors that the parsing host might return (application errors etc)

Libraries.io currently relies on those errors being swallowed here so this will be a breaking change and major version bump, as @katzj we should make it consistent across all parsers that call out to external services too

andrew:
Rebasing this branch against master should fix the Tidelift CVE error: https://github.com/librariesio/bibliothecary/commit/adc1999d3e5f6c406388b7cd1f6735177903e519

PR #425, opened 2018-09-10
gsub would substitute all instances of cwd i.e. '.' so switch to sub …

mellisio:
…instead

### DIFF
--- a/lib/bibliothecary.rb
+++ b/lib/bibliothecary.rb
@@ -10,17 +10,12 @@ end
 
 module Bibliothecary
   def self.analyse(path)
-    file_list = load_file_list(path)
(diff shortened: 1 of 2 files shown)


lib/bibliothecary.rb
```diff
@@ -10,12 +10,17 @@
 
 module Bibliothecary
   def self.analyse(path)
+    file_list = load_file_list(path)
+    package_managers.map{|pm| pm.analyse(path, file_list) }.flatten.compact
+  end
+
+  def self.load_file_list(path)
```
katzj: This refactor seems fine but is unrelated? Should we now do better testing of `load_file_test`?

mellisio: I split that out to enable better testing of it but I didn't write the tests yet. I can include them with this PR if that's better.

PR #426, opened 2018-09-10
Add a success flag so that we can tell if we successfully parsed a ma…

mellisio:
…nifest file

### DIFF
--- a/lib/bibliothecary/analyser.rb
+++ b/lib/bibliothecary/analyser.rb
@@ -100,11 +100,18 @@ module Bibliothecary
             platform: platform_name,
             path: filename,
             dependencies: dependencies,
-            kind: determine_kind(filename)
+            kind: determine_kind(filename),
+            success: true
           }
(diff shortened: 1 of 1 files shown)


lib/bibliothecary/analyser.rb
```diff
@@ -100,11 +100,18 @@ def analyse_contents(filename, contents)
...
           }
         end
-      # rescue
-      #   nil
+      rescue
+        {
+          platform: platform_name,
+          path: filename,
+          dependencies: nil,
+          kind: determine_kind(filename),
```
katzj: Can this throw an error? Actually looking below, it doesn't look like it but the case where nothing matches seems like it's weird as kind is a string and not an array type?

mellisio: So kind is a string in the success case? I'm not really certain what we want to happen there I'll dig into that a little more.

PR #446, opened 2019-04-01
Don't use regexes while matching paths

kbarrette:
I'm able to get around a 5x speedup by ditching regexes for matching paths.

I've left the original regexes in place for comparison, I'll remove them before this merges.

The first commit fixes a spec that's failing on master (at least for me locally)

### DIFF
--- a/lib/bibliothecary/parsers/hackage.rb
+++ b/lib/bibliothecary/parsers/hackage.rb
@@ -8,11 +8,11 @@ module Bibliothecary
 
       def self.mapping
         {
-          /.*\.cabal$/ => {
+          match_extension(".cabal") => {
             kind: 'manifest',
             parser: :parse_cabal
           },
-          /^cabal\.config$|.*\/cabal\.config$/ => {
+          match_extension("cabal.config") => {
             kind: 'lockfile',
             parser: :parse_cabal_config
           },
--- a/lib/bibliothecary/parsers/pypi.rb
+++ b/lib/bibliothecary/parsers/pypi.rb
@@ -6,24 +6,25 @@ module Bibliothecary
       INSTALL_REGEXP = /install_requires\s*=\s*\[([\s\S]*?)\]/
       REQUIRE_REGEXP = /([a-zA-Z0-9]+[a-zA-Z0-9\-_\.]+)([><=\w\.,]+)?/
       REQUIREMENTS_REGEXP = /^#{REQUIRE_REGEXP}/
+      MANIFEST_REGEXP = /.*require[^\/]*(\/)?[^\/]*\.(txt|pip)$/
 
       def self.mapping
         {
-          /.*require[^\/]*(\/)?[^\/]*\.(txt|pip)$/ => {
+          lambda { |p| MANIFEST_REGEXP.match(p) } => {
             kind: 'manifest',
             parser: :parse_requirements_txt,
             can_have_lockfile: false
           },
-          /^setup\.py$|.*\/setup.py$/ => {
+          match_filename("setup.py") => {
             kind: 'manifest',
             parser: :parse_setup_py,
             can_have_lockfile: false
           },
-          /^Pipfile$|.*\/Pipfile$/ => {
+          match_filename("Pipfile") => {
             kind: 'manifest',
             parser: :parse_pipfile
           },
-          /^Pipfile\.lock$|.*\/Pipfile\.lock$/ => {
+          match_filename("Pipfile.lock") => {
             kind: 'lockfile',
             parser: :parse_pipfile_lock
           }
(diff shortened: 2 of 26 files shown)


lib/bibliothecary/parsers/rubygems.rb
```diff
@@ -10,16 +10,19 @@ class Rubygems
 
       def self.mapping
         {
-          /^Gemfile$|^gems\.rb$|.*\/Gemfile$|.*\/gems\.rb$/ => {
+          # /^Gemfile$|^gems\.rb$|.*\/Gemfile$|.*\/gems\.rb$/ => {
+          lambda { |p| match_filename("Gemfile")[p] || match_filename("gems.rb")[p] } => {
```
jsonperl: Could add a match_filenames, or take a list and [filenames].flatten it?

jsonperl: I stared at the [p]s for minutes trying to figure out what was happening. I had no idea you could use lambdas like that!

spec/parsers/rubygems_spec.rb
```diff
@@ -52,7 +52,7 @@
         {:name=>"orm_adapter", :requirement=>"~> 0.1", :type=>:development},
         {:name=>"bcrypt", :requirement=>"~> 3.0", :type=>:runtime},
         {:name=>"thread_safe", :requirement=>"~> 0.1", :type=>:runtime},
-        {:name=>"railties", :requirement=>">= 3.2.6, < 5", :type=>:runtime},
```
jsonperl: What's that about?

kbarrette: No idea, it's failing on master though. Would love advice from someone who has more context here.

lib/bibliothecary/parsers/rubygems.rb
kbarrette: Yeah, I wasn't sure about making `match_filenames` for just these few cases

spec/parsers/rubygems_spec.rb
kbarrette: As far as I can tell, this is due to https://github.com/ruby/ruby/commit/8da8d4b043c37b53a69803c71ff36b478d4776d0, which is NOT in Ruby 2.5.1, but IS in whatever Travis is calling Ruby 2.5.1

kbarrette: @jsonperl @katzj I added consistent sorting of requirements

jsonperl:
Oh @kbarrette probably bump the version.rb, and then we can tag on merge.

kbarrette:
@jsonperl went with `6.4.0`, unless we think people are adding their own stuff in `Bibliothecary::Parsers`, in which case this might be a major version bump.

jsonperl:
SHIP IT

lib/bibliothecary/parsers/rubygems.rb
katzj: I think it's worth adding `match_filenames` for clarity even though only used a few times

lib/bibliothecary/parsers/hackage.rb
```diff
@@ -8,11 +8,11 @@ class Hackage
 
       def self.mapping
         {
-          /.*\.cabal$/ => {
+          lambda { |p| p.end_with?(".cabal") } => {
```
katzj: I wonder if it's also worth having `filename_endswith` as that's used a lot

lib/bibliothecary/parsers/pypi.rb
```diff
@@ -6,24 +6,25 @@ class Pypi
       INSTALL_REGEXP = /install_requires\s*=\s*\[([\s\S]*?)\]/
       REQUIRE_REGEXP = /([a-zA-Z0-9]+[a-zA-Z0-9\-_\.]+)([><=\w\.,]+)?/
       REQUIREMENTS_REGEXP = /^#{REQUIRE_REGEXP}/
+      MANIFEST_REGEXP = /.*require[^\/]*(\/)?[^\/]*\.(txt|pip)$/
 
       def self.mapping
         {
-          /.*require[^\/]*(\/)?[^\/]*\.(txt|pip)$/ => {
+          lambda { |p| MANIFEST_REGEXP.match(p) } => {
```
katzj: And `match_regexp`

kbarrette:
I'd been considering making a `Matcher` class - let me see how that looks.

kbarrette:
@jsonperl @katzj after experimenting with this a bit, the temptation is to go one of two ways:

1. Create some kind of builder or DSL like `FilenameMatcher.match("pom.xml").case_insensitive`
2. Write specific methods for each situation: `match_filename`, `match_filenames`, `match_case_insensitive_filename`, `match_regexp`, `match_ends_with`, `match_case_insensitive_ends_with`, etc.

I think I like my original implementation best out of the three, although I could be convinced to do something else if people felt strongly about it.

PR #452, opened 2019-05-03
Update rubocop yaml

tiegz:
* updates Rubocop rules
* runs `rubocop -a` to autofix rule violations


.rubocop.yml
```diff
@@ -257,62 +240,6 @@ Metrics/PerceivedComplexity:
                  human reader.
   Enabled: false
 
-##################### Performance #############################
```
tiegz: The Performance rules were extracted into a separate gem, and the logic around Ruby version differences makes sense, so this just removes the rules.

PR #453, opened 2019-05-05
Adding parser support for Go Modules

tiegz:
Kicking this off. 

### Docs

* [Go Modules wiki](https://github.com/golang/go/wiki/Modules)
* [Go Modules proposal](https://go.googlesource.com/proposal/+/master/design/24301-versioned-go.md)
* [good intro to Go Modules](https://roberto.selbach.ca/intro-to-go-modules/)

### Example Go Modules package

* [go-yaml/yaml on GH](https://github.com/go-yaml/yaml)
* [go-yaml/yaml on go-search](https://go-search.org/view?id=gopkg.in%2fyaml.v2)
* [go-yaml/yaml on Libraries, notably missing deps](https://libraries.io/go/github.com%2Fgo-yaml%2Fyaml)

### Open Questions

* [x] should we maintain the full [pseudo-version](https://golang.org/cmd/go/#hdr-Pseudo_versions) for packages that don't have semver?
* [x] good non-GH examples?
* [x] I've marked `go.sum` as a `checksum` file because the docs [explicitly say it isn't a lock file](https://github.com/golang/go/wiki/Modules#is-gosum-a-lock-file-why-does-gosum-include-information-for-module-versions-i-am-no-longer-using). But it does break down indirect deps and can have more than `go.sum`, so maybe we **should** consider it a `lockfile`?
* [x] these regexps feel a bit janky? But Go Modules notably chose a non-standard depfile syntax (see [this discussion](https://github.com/golang/go/issues/23966))

### DIFF
--- a/lib/bibliothecary/parsers/go.rb
+++ b/lib/bibliothecary/parsers/go.rb
@@ -124,14 +124,13 @@ module Bibliothecary
       def self.parse_go_sum(file_contents)
         deps = []
         file_contents.lines.map(&:strip).each do |line|
-          match = line.match(GOSUM_REGEX)
-
-          next unless match
-          deps << {
-            name: match[1].strip,
-            requirement: match[2].strip.split('/').first || '*',
-            type: 'runtime'
-          }
+          if match = line.match(GOSUM_REGEX)
+            deps << {
+              name: match[1].strip,
+              requirement: match[2].strip.split('/').first || '*',
+              type: 'runtime'
+            }
+          end
         end
         deps.uniq
       end
--- a/lib/bibliothecary/parsers/go.rb
+++ b/lib/bibliothecary/parsers/go.rb
@@ -7,9 +7,19 @@ module Bibliothecary
       include Bibliothecary::Analyser
 
       GPM_REGEXP = /^(.+)\s+(.+)$/
+      GOMOD_REGEX = /^(.+)\s+(.+)$/
+      GOSUM_REGEX = /^(.+)\s+(.+)\s+(.+)$/
 
       def self.mapping
         {
+          match_filename("go.mod") => {
+            kind: 'manifest',
+            parser: :parse_go_mod
+          },
+          match_filename("go.sum") => {
+            kind: 'checksum',
+            parser: :parse_go_sum
+          },
           match_filename("glide.yaml") => {
             kind: 'manifest',
             parser: :parse_glide_yaml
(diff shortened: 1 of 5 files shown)


lib/bibliothecary/parsers/go.rb
```diff
@@ -7,9 +7,19 @@ class Go
       include Bibliothecary::Analyser
 
       GPM_REGEXP = /^(.+)\s+(.+)$/
+      GOMOD_REGEX = /^(.+)\s+(.+)$/
+      GOSUM_REGEX = /^(.+)\s+(.+)\s+(.+)$/
```
tiegz: 3rd capture here is a checksum of the dep, which I don't think we need.

lib/bibliothecary/parsers/go.rb
```diff
@@ -95,6 +105,39 @@ def self.parse_dep_lockfile(file_contents)
         map_dependencies(manifest, 'projects', 'name', 'revision', 'runtime')
       end
 
+      def self.parse_go_mod(file_contents)
+        deps = []
+        file_contents.lines.map(&:strip).each do |line|
+          next if line.start_with?("module ")
```
jsonperl: Any interest in regex capturing this?
```
lines = file_contents.match(/require\s+\(\s?(.*)\)/m).captures[0]
```

jsonperl: I'll just put this here before someone beats me to it.
Screen Shot 2019-05-06 at 9 24 18 AM

lib/bibliothecary/parsers/go.rb
```diff
@@ -95,6 +105,39 @@ def self.parse_dep_lockfile(file_contents)
...
+          next if line.start_with?("go ")
+          next if line.start_with?("require (")
+          next if line.start_with?(")")
+          match = line.gsub(/(\/\/(.*))/, '').match(GOMOD_REGEX)
```
jsonperl: Is this for the comment stuff at the end?

jsonperl:
Can try some gopkg.in urls as some non-github examples? I don't know if it's a thing anymore but I've got projects that used it. Supposedly it's a way to mark a stable version for a github repository, and I'm not sure how that gels with modules. But folks probably still have em in their projects.

ex: gopkg.in/gin-gonic/gin.v1

jsonperl:
Re psuedo-versions, is the idea generally that it's a fallback for stuff that hasn't yet opted into modules? (I skimmed the link but I'm still not fully grokking it)

lib/bibliothecary/parsers/go.rb
```diff
@@ -7,9 +7,19 @@ class Go
...
+      GOSUM_REGEX = /^(.+)\s+(.+)\s+(.+)$/
 
       def self.mapping
         {
+          match_filename("go.mod") => {
+            kind: 'manifest',
+            parser: :parse_go_mod
+          },
+          match_filename("go.sum") => {
+            kind: 'checksum',
```
katzj: A new file `kind` will have some implications for callers/users

tiegz: I think `lockfile` might make sense after all: it has packages that don't show up in `go.mod`, and technically a checksum is like an additional lock on the version 🤷‍♂

lib/bibliothecary/parsers/go.rb
tiegz: Yeah, there's an `// indirect` that gets added sometimes, and I'm not sure yet if there are other comments besides `indirect` so I just did a catch-all. Here's a note about it ([ref](https://github.com/golang/go/wiki/Modules#faqs--gomod-and-gosum)):

```If a dependency of your module does not itself have a go.mod (e.g., because the dependency has not yet opted in to modules itself), or if its go.mod file is missing one or more of its dependencies (e.g., because the module author did not run go mod tidy), then the missing transitive dependencies will be added to your module's requirements, along with an // indirect comment to indicate that the dependency is not from a direct import within your module.```

tiegz:
@jsonperl thanks, I hadn't used gopkg.in before, so added that!

> Re psuedo-versions, is the idea generally that it's a fallback for stuff that hasn't yet opted into modules?

Yeah, exactly, these are applied when there are no semver tags available in the repo (so it attaches the "0.0.0" version, plus the data and sha of the commit).

PR #454, opened 2019-05-29
Rubygems parser Gemfile.lock chomp nonprinting newline chars

tyrelsouza:
We tried scanning a file that appears to have been made on windows. This caused there to be an error message of: `Gemfile.lock: undefined method 'gsub' for nil:NilClass"` because the match didn't match ` name (ver)\r`

Deciding to chomp away the newline characters

### DIFF
--- a/lib/bibliothecary/parsers/rubygems.rb
+++ b/lib/bibliothecary/parsers/rubygems.rb
@@ -29,7 +29,8 @@ module Bibliothecary
       end
 
       def self.parse_gemfile_lock(manifest)
-        manifest.split("\n").map do |line|
+        manifest.each_line.map do |line|
+          line.chomp!
           match = line.match(NAME_VERSION_4)
           next unless match
           name = match[1]
--- a/lib/bibliothecary/parsers/rubygems.rb
+++ b/lib/bibliothecary/parsers/rubygems.rb
@@ -29,8 +29,7 @@ module Bibliothecary
       end
 
       def self.parse_gemfile_lock(manifest)
-        manifest.each_line.map do |line|
-          line.chomp!
+        manifest.lines(chomp: true).map do |line|
           match = line.match(NAME_VERSION_4)
           next unless match
           name = match[1]
--- a/spec/parsers/rubygems_spec.rb
+++ b/spec/parsers/rubygems_spec.rb
@@ -76,10 +76,15 @@ describe Bibliothecary::Parsers::Rubygems do
   end
 
   it 'parses dependencies from Gemfile.lock with windows line endings' do
+    fixture = load_fixture("GemfileLineEndings.lock")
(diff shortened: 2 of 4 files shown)


lib/bibliothecary/parsers/rubygems.rb
```diff
@@ -30,10 +30,11 @@ def self.mapping
 
       def self.parse_gemfile_lock(manifest)
         manifest.split("\n").map do |line|
```
jsonperl: I think we can rely on ruby's https://apidock.com/ruby/String/each_line. Maybe give that a shot and run a couple scenarios through it to make sure it's true?

tyrelsouza: Seems that doesn't actually strip the chars though, still need to remove them
```
irb(main):002:0> "alksdjf\nlaksdjf\r\nlaksjdf\nlkajsdf\rlkjasdlfkj".each_line { |l| puts l.dump }
"alksdjf\n"
"laksdjf\r\n"
"laksjdf\n"
"lkajsdf\rlkjasdlfkj"
=> "alksdjf\nlaksdjf\r\nlaksjdf\nlkajsdf\rlkjasdlfkj"
```

lib/bibliothecary/parsers/rubygems.rb
```diff
@@ -29,7 +29,7 @@ def self.mapping
       end
 
       def self.parse_gemfile_lock(manifest)
-        manifest.split("\n").map do |line|
+        manifest.each_line.map(&:chomp!).map do |line|
```
jsonperl: We can probably ditch the second map, and probably with doing in the first line of the block so as not to pre-create the array.
```
manifest.each_line.map do |line|
  line = line.chomp!
  ...
end
```

spec/parsers/rubygems_spec.rb
```diff
@@ -75,6 +75,21 @@
     })
   end
 
+  it 'parses dependencies from Gemfile.lock with windows line endings' do
+    expect(
+      described_class.analyse_contents(
+        "Gemfile.lock",
```
katzj: Maybe throw this into a fixture?

lib/bibliothecary/parsers/rubygems.rb
```diff
@@ -29,7 +29,8 @@ def self.mapping
       end
 
       def self.parse_gemfile_lock(manifest)
-        manifest.split("\n").map do |line|
+        manifest.each_line.map do |line|
+          line.chomp!
```
tiegz:
```suggestion:-1
        manifest.lines(chomp: true).map do |line|
```

jsonperl: Ooh yess!

PR #455, opened 2019-05-29
Adds Bibliothecary.configuration.ignored_deps option

tiegz:
### Usage

``` ruby
    Bibliothecary.configure do |config|
      config.ignored_deps = {
        "composer.json" => [/^php$/i]
      }
    end
```

This will omit any dependencies from `"composer.json"` files that matches the regexes specified.

### Why?

You may want to skip certain dependencies from an analysis. For example, Composer [allows you](https://getcomposer.org/doc/04-schema.md#package-links) to specify the version of PHP or the required system extensions for PHP alongside your regular deps:

``` json
"require": {
		"laravel/framework": "5.0.*"
		"laravel/framework": "5.0.*",
		"php": "^7.1",
		"ext-mbstring": "*"
	},
```

But `php` (the language) and `ext-mbstring` (a system extension) aren't really packages, so this option lets you skip them when parsing a manifest.

### DIFF
--- a/lib/bibliothecary/configuration.rb
+++ b/lib/bibliothecary/configuration.rb
@@ -14,7 +14,7 @@ module Bibliothecary
     def initialize
       @ignored_dirs = ['.git', 'node_modules', 'bower_components', 'vendor', 'dist']
       @ignored_files = []
-      @ignored_deps = {} # e.g. { 'packagist' => [/a_regexp/] }
+      @ignored_deps = {} # e.g. { 'composer.json' => [/some_depname_regexp/, /another_depname_regexp/] }
       @carthage_parser_host = 'https://carthage.libraries.io'
       @clojars_parser_host  = 'https://clojars.libraries.io'
       @mix_parser_host      = 'https://mix.libraries.io'
--- a/lib/bibliothecary/configuration.rb
+++ b/lib/bibliothecary/configuration.rb
@@ -2,6 +2,7 @@ module Bibliothecary
   class Configuration
     attr_accessor :ignored_dirs
     attr_accessor :ignored_files
+    attr_accessor :ignored_deps
     attr_accessor :carthage_parser_host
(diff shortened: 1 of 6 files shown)


spec/bibliothecary_spec.rb
```diff
@@ -282,22 +305,18 @@
...
   end
 
-  it 'allows customization of config options' do
-    Bibliothecary.configure do |config|
-      config.ignored_dirs = ['foobar']
-    end
-
-    expect(Bibliothecary.ignored_dirs).to eq(['foobar'])
-  end
-
```
tiegz: This was a dupe test

lib/bibliothecary/configuration.rb
```diff
@@ -13,6 +14,7 @@ class Configuration
     def initialize
       @ignored_dirs = ['.git', 'node_modules', 'bower_components', 'vendor', 'dist']
       @ignored_files = []
+      @ignored_deps = {} # e.g. { 'packagist' => [/a_regexp/] }
```
jsonperl: Seems like you changed gears since the comment

jsonperl: ☝️ Change gears = It was a platform => regex hash, and is now keyed on file.

tiegz:
@jsonperl this is mainly to work around package managers that mix deps with non-deps in their manifests. Composer does this, but I'm wondering if this will be so rare that it doesn't necessitate this option. Is this work better done by clients of the library? cc @katzj

jsonperl:
Gotcha @tiegz thanks for the info. I don't see any harm in trying it out here, and seeing how it plays out. We can always pull it back out into a client if it doesn't prove useful.

katzj:
I think we probably want to keep it in the caller side... there's a level of options that just start to get hard to understand if we natively have it here.

PR #457, opened 2019-06-06
Parse Maven Dependency Plugin Output

mikeyoung85:
Support a file containing resolved maven dependencies from maven dependency plugin and treat it like a Maven lockfile.


lib/bibliothecary/parsers/maven.rb
```diff
@@ -11,6 +11,8 @@ class Maven
       # "|    \\--- com.google.guava:guava:23.5-jre (*)"
       GRADLE_DEP_REGEX = /(\+---|\\---){1}/
 
+      MAVEN_DEP_REGEXT = /(\s+)([\w\.-]+)\:([\w\.-]+)\:([\w\.-]+)\:([\w\.-]+)\:([\w]+).+/
```
mikeyoung85: My regex fu is weak but this is meant to match up each section of the dependency string so that can easily be set in the correct location in the parser.

lib/bibliothecary/parsers/maven.rb
```diff
@@ -109,6 +115,20 @@ def self.parse_gradle_resolved(file_contents)
...
+          type_match = MAVEN_DEP_REGEXT.match(line)
+
+          next unless type_match
+          # org.springframework.boot:spring-boot-starter-web:jar:2.0.3.RELEASE:compile[36m -- module spring.boot.starter.web[0;1m [auto][m
+          {
+            name: type_match[2, 2].join(":"),
+            requirement: type_match[5],
+            type: type_match[6]
+          }
+        end.compact.uniq {|item| [item[:name], item[:requirement], item[:type]]}
```
kbarrette: I really don't like the look of chaining with `do`/`end`, so I generally make another method, for something like:
```ruby
file_contents
  .lines
  .map(&method(:parse_line))
  .compact
  .uniq { ... }
```

lib/bibliothecary/parsers/maven.rb
```diff
@@ -109,6 +115,20 @@ def self.parse_gradle_resolved(file_contents)
         end.compact.uniq {|item| [item[:name], item[:requirement], item[:type]]}
       end
 
+      def self.parse_maven_resolved(file_contents)
+        file_contents.split("\n").map do |line|
+          type_match = MAVEN_DEP_REGEXT.match(line)
```
kbarrette: Can we just split on `:` here? It might be easier to reason about than the regex

mikeyoung85: Yeah that sounds more reasonable.

lib/bibliothecary/parsers/maven.rb
```diff
@@ -109,6 +113,25 @@ def self.parse_gradle_resolved(file_contents)
...
+      end
+
+      def self.parse_resolved_dep_line(line)
+        dep_parts = line.strip.split(":")
+        return nil unless dep_parts.length == 5
+        # org.springframework.boot:spring-boot-starter-web:jar:2.0.3.RELEASE:compile[36m -- module spring.boot.starter.web[0;1m [auto][m
+        {
+          name: dep_parts[0, 2].join(":"),
+          requirement: dep_parts[3],
+          type: dep_parts[4].split("--").first.gsub(/\e\[(\d+)m/, '').strip #remove control characters from string
```
mikeyoung85: I think these are ANSI color codes that have shown up in the output. I think I'll need to take a deeper look at this because this could become a problem depending on the environment the maven dependency file was generated in.

mikeyoung85: Added a more comprehensive regex sub.

lib/bibliothecary/parsers/maven.rb
```diff
@@ -11,6 +11,8 @@ class Maven
       # "|    \\--- com.google.guava:guava:23.5-jre (*)"
       GRADLE_DEP_REGEX = /(\+---|\\---){1}/
 
+      ANSI_MATCHER = '(\[)?\033(\[)?[;?\d]*[\dA-Za-z]([\];])?'
```
mikeyoung85: This should match the ANSI codes.

lib/bibliothecary/parsers/maven.rb
```diff
@@ -109,6 +115,26 @@ def self.parse_gradle_resolved(file_contents)
         end.compact.uniq {|item| [item[:name], item[:requirement], item[:type]]}
       end
 
+      def self.parse_maven_resolved(file_contents)
+        file_contents
+          .gsub(/#{ANSI_MATCHER}/, '')
```
mikeyoung85: Use `gsub` to remove ANSI characters from the file string.

lib/bibliothecary/parsers/maven.rb
kbarrette: I wonder if we should use a gem here? https://github.com/fazibear/colorize ?

lib/bibliothecary/parsers/maven.rb
kbarrette: You can make the constant the regex itself, and not have to do this interpolation

lib/bibliothecary/parsers/maven.rb
mikeyoung85: I went with https://github.com/piotrmurach/strings-ansi to handle this instead. I had originally pulled the regex from this project and it seems to work where the `colorize` gem wasn't picking up all the codes.

lib/bibliothecary/parsers/maven.rb
```diff
@@ -109,6 +114,26 @@ def self.parse_gradle_resolved(file_contents)
         end.compact.uniq {|item| [item[:name], item[:requirement], item[:type]]}
       end
 
+      def self.parse_maven_resolved(file_contents)
+        file_contents = Strings::ANSI.sanitize(file_contents)
```
kbarrette: You can skip the variable and just
```ruby
Strings::ANSI
  .sanitize(file_contents)
  .lines
  .map ...
```

lib/bibliothecary/parsers/maven.rb
```diff
@@ -109,6 +114,26 @@ def self.parse_gradle_resolved(file_contents)
...
+        file_contents
+          .split("\n")
+          .map(&method(:parse_resolved_dep_line))
+          .compact
+          .uniq {|item| [item[:name], item[:requirement], item[:type]]}
```
kbarrette: You shouldn't need to specify what to uniq on, since it's the entire hash

lib/bibliothecary/parsers/maven.rb
```diff
@@ -109,6 +114,26 @@ def self.parse_gradle_resolved(file_contents)
...
+      end
+
+      def self.parse_resolved_dep_line(line)
+        dep_parts = line.strip.split(":")
+        return nil unless dep_parts.length == 5
```
kbarrette: Convention would normally be just `return unless` (no `nil`)

PR #458, opened 2019-07-01
More Maven Property Resolving

mikeyoung85:
This will resolve properties that are hanging out inside of a string like `http://www.myhomepage.com/${project.artifactId}`. It will also recurse up to 5 times to resolve a property. Added some changes to the spec pom file to verify the resolved values.

### DIFF
--- a/lib/bibliothecary/parsers/maven.rb
+++ b/lib/bibliothecary/parsers/maven.rb
@@ -13,6 +13,7 @@ module Bibliothecary
       GRADLE_DEP_REGEX = /(\+---|\\---){1}/
 
       MAVEN_PROPERTY_REGEX = /^(.*)\$\{(.+)\}(.*)/
+      MAX_DEPTH = 5
 
       def self.mapping
         {
@@ -189,9 +190,22 @@ module Bibliothecary
       end
 
       def self.extract_property(xml, property_name, value, parent_properties = {}, depth = 0)
+        prop_value = property_value(xml, property_name, parent_properties)
+        return value unless prop_value
         # don't resolve more than 5 levels deep to avoid potential circular references
-        max_depth = 5
+       
+        resolved_value = replace_value_with_prop(value, prop_value, property_name)
+        # check to see if we just resolved to another property name
+        match = resolved_value.match(MAVEN_PROPERTY_REGEX)
+        if match && depth < MAX_DEPTH
+          depth += 1
+          return extract_property(xml, match[2], resolved_value, parent_properties, depth)
+        else
+          return resolved_value
+        end 
+      end
 
+      def self.property_value(xml, property_name, parent_properties)
         # the xml root is <project> so lookup the non property name in the xml
         # this converts ${project/group.id} -> ${group/id}
         non_prop_name = property_name.gsub(".", "/").gsub("project/", "")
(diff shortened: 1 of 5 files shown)


lib/bibliothecary/parsers/maven.rb
```diff
@@ -172,30 +174,53 @@ def self.extract_pom_info(xml, location, parent_properties = {})
...
-          elsif parent_prop
-            return parent_prop
-          elsif xml.locate(non_prop_name).first
-            # see if the value to look up is a field under the project
-            # examples are ${project.groupId} or ${project.version}
-            return xml.locate(non_prop_name).first.nodes.first
-          else
-            return value
-          end
+          return extract_property(xml, match[2], value, parent_properties)
```
kbarrette: If we only ever care about the second capture, we can ditch the first.

lib/bibliothecary/parsers/maven.rb
```diff
@@ -12,6 +12,8 @@ class Maven
       # "|    \\--- com.google.guava:guava:23.5-jre (*)"
       GRADLE_DEP_REGEX = /(\+---|\\---){1}/
 
+      MAVEN_PROPERTY_REGEX = /^(.*)\$\{(.+)\}(.*)/
```
kbarrette: I think `/\$/{(.+?)\}/` would do here

lib/bibliothecary/parsers/maven.rb
```diff
@@ -172,30 +174,53 @@ def self.extract_pom_info(xml, location, parent_properties = {})
...
         else
           return value
         end
       end
+
+      def self.replace_value_with_prop(original_value, property_value, property_name)
+        original_value.gsub("${#{property_name}}", property_value)
+      end
+
+      def self.extract_property(xml, property_name, value, parent_properties = {}, depth = 0)
```
kbarrette: I think if you break this into two methods it'll flow more easily, something like:
```ruby
      def self.extract_property(xml, property_name, value, parent_properties = {}, depth = 0)
        prop_value = property_value(xml, property_name, parent_properties)
        return value unless prop_value
        resolved_value = replace_value_with_prop(value, prop_value, property_name)

        # check to see if we just resolved to another property name
        match = resolved_value.match(MAVEN_PROPERTY_REGEX)
        if match && depth < MAX_DEPTH
          extract_property(xml, match[2], resolved_value, parent_properties, depth + 1)
        else
          resolved_value
        end
      end

      def self.property_value(xml, property_name, parent_properties)
        # the xml root is <project> so lookup the non property name in the xml
        # this converts ${project/group.id} -> ${group/id}
        prop_field = xml.respond_to?(:properties) && xml.properties.locate(property_name).first
        non_prop_name = property_name.gsub(".", "/").gsub("project/", "")

        if prop_field
          prop_field.nodes.first
        elsif parent_properties[property_name]
          parent_properties[property_name]
        elsif (non_prop = xml.locate(non_prop_name).first)
          # see if the value to look up is a field under the project
          # examples are ${project.groupId} or ${project.version}
          non_prop.nodes.first
        end

        # property not found
      end
```

lib/bibliothecary/parsers/maven.rb
mikeyoung85: It wouldn't capture any properties in the middle of the string, which was one of the problems we were having.

lib/bibliothecary/parsers/maven.rb
kbarrette: n.b. I did not test this

lib/bibliothecary/parsers/maven.rb
kbarrette: That was due to the `^` marking "beginning of line"

PR #461, opened 2019-08-08
Initial draft of putting Conda onto Bibliothecary

tyrelsouza:


lib/bibliothecary/parsers/conda.rb
```diff
@@ -0,0 +1,38 @@
...
+      end
+
+      def self.parse_yaml_manifest(file_contents)
+        unparsed_manifest = YAML.load file_contents
+
+        manifest = {}
+        manifest["dependencies"] = unparsed_manifest["dependencies"].map do |dep|
+          # fields seem to be  `name=version=build` or `name` or `name=version`
+          # Handling the `name` case by always appending a ">= 0" to the list, 
+          # and taking first two items.
```
tyrelsouza: haven't seen any online where it's something like 

`name=build`  so not sure if we should handle that. It seems people either hand craft in

`name`
`name=version`
or automate it export being
`name=version=build`

lib/bibliothecary/configuration.rb
```diff
@@ -18,6 +19,7 @@ def initialize
       @mix_parser_host      = 'https://mix.libraries.io'
       @gradle_parser_host   = 'https://gradle-parser.libraries.io'
       @yarn_parser_host     = 'https://yarn-parser.libraries.io'
+      @conda_parser_host    = 'https://conda.libraries.io'
```
tyrelsouza: Working on getting some of these timeout errors to go away - so this can't merge until some internal things are settled.

lib/bibliothecary/parsers/conda.rb
```diff
@@ -0,0 +1,60 @@
...
+module Bibliothecary
+  module Parsers
+    class Conda
+      include Bibliothecary::Analyser
+      FILE_KINDS = %w[manifest lockfile]
+
+      def self.mapping
+        {
+          match_filename("environment.yml") => {
+            parser: :parse_conda_environment,
```
kbarrette: I don't think we're using `parser` at all for conda? If not, let's make it a no-op

lib/bibliothecary/analyser.rb
```diff
@@ -89,12 +89,11 @@ def parse_json_runtime_manifest(file_contents)
       end
 
       def map_dependencies(hash, key, type)
-        hash.fetch(key,[]).map do |name, requirement|
-          {
-            name: name,
-            requirement: requirement,
-            type: type
-          }
+        # determine if the items are hashes or arrays, and return appropriate results
```
kbarrette: I think we can keep this internal to `Conda` by removing this change and just having `Conda` do the right thing

lib/bibliothecary/parsers/conda.rb
```diff
@@ -0,0 +1,60 @@
...
+        }
+      end
+
+      def self.parse_conda_environment(file_contents)
+        manifest = call_conda_parser_web(file_contents)
+
+        Hash[FILE_KINDS.collect { |kind| [kind, map_dependencies(manifest, kind, "runtime")] }]
+      end
+
+      def self.call_conda_parser_web(file_contents)
```
kbarrette: Looks like this can be private

lib/bibliothecary/parsers/conda.rb
```diff
@@ -0,0 +1,60 @@
...
+        host = Bibliothecary.configuration.conda_parser_host
+        response = Typhoeus.post(
+          "#{host}/parse",
+          headers: {
+              ContentType: 'multipart/form-data'
+          },
+          # TODO: Can we ever get the filename from the method call from `match_filename` in the future? would be nice to be able to.
```
kbarrette: Not sure I understand this TODO

lib/bibliothecary/analyser.rb
tyrelsouza: I kind of disagree. Yes there's the YAGNI principle, but keeping this in will allow future parsers to not care whether they are a two-tuple of `(name, requirement)` or a nicely formatted response from a json-api.

(for instance, with this in, I can use it in `npm.rb`, `hackage.rb`, `hex.rb`, `maven.rb`, etc..  as well, rather than having  almost the same maps to maintain there.
I might do this in a new PR after this, I think it's out of scope of this PR, but will be a small change)

lib/bibliothecary/parsers/conda.rb
tyrelsouza: Updated it, and added a higher up TODO for more context!

PR #465, opened 2019-09-26
Allow matching from multiple kinds

tyrelsouza:


lib/bibliothecary/parsers/conda.rb
```diff
@@ -4,7 +4,7 @@ module Bibliothecary
   module Parsers
     class Conda
       include Bibliothecary::Analyser
-      FILE_KINDS = %i[manifest lockfile]
+      FILE_KINDS = %w[manifest lockfile]
```
tyrelsouza: Making this string now is the big clincher

PR #467, opened 2019-09-30
Make conda return Lockfile, or Manifest, not both in one response.

tyrelsouza:
conda-parser change is live, so updated VCR records.
of course new versions have released, so the responses are _slightly_ different with versions.

Also only returning one type, so NO MORE HARDCODING making a `Bibliothecary::Analyser.create_analysis(`  i can just return json!

For more context, see https://github.com/librariesio/conda-parser/pull/16


spec/parsers/conda_spec.rb
```diff
@@ -17,116 +17,98 @@
           {:name=>"ncurses", :requirement=>"6.1", :type=>"runtime"},
           {:name=>"numpy", :requirement=>"1.16.4", :type=>"runtime"},
           {:name=>"openssl", :requirement=>"1.1.1c", :type=>"runtime"},
-          {:name=>"pip", :requirement=>"19.2.2", :type=>"runtime"},
```
tyrelsouza: no version because we're not resolving the version now for manifests, just returning as is. this example file has pip and setuptools without versions

lib/bibliothecary/parsers/conda.rb
```diff
@@ -1,67 +1,53 @@
...
+          },
+          match_filename("environment.yaml.lock") => {
+            parser: :parse_conda_lockfile,
+            kind: "lockfile",
+          },
         }
       end
 
-      # Overrides Analyser.analyse_contents_from_info
-      def self.analyse_contents_from_info(info)
```
tyrelsouza: GOOD BYE OVERRIDE!

lib/bibliothecary/parsers/conda.rb
```diff
@@ -1,67 +1,53 @@
...
-      rescue Psych::SyntaxError => e
-        Bibliothecary::Analyser::create_error_analysis(platform_name, info.relative_path, "runtime", e.message)
-      end
-
-      private
-
       def self.parse_conda(info)
-        results = call_conda_parser_web(info.contents)
-        FILE_KINDS.map do |kind|
-          Bibliothecary::Analyser.create_analysis(
```
tyrelsouza: SEE YA Explicit Creation

PR #472, opened 2020-05-13
add resolved pip to pypi

tyrelsouza:


lib/bibliothecary/parsers/pypi.rb
```diff
@@ -15,6 +15,11 @@ def self.mapping
             parser: :parse_requirements_txt,
             can_have_lockfile: false
           },
+          match_filename('pip-resolved-dependencies.txt') => { # Inferred from pip
+            kind: 'lockfile',
+            parser: :parse_requirements_txt,
+            can_have_lockfile: false
```
tyrelsouza, line 21: should this be true, it IS a lockfile...

PR #474, opened 2020-06-06
Handle single-line require expressions in go.mod.

tiegz:
Buffalo's events module has an example of this case: https://github.com/gobuffalo/events/blob/master/go.mod

Also, adds ignoring of `exclude` and `replace` directives, since Bibliothecary doesn't need those.


lib/bibliothecary/parsers/go.rb
```diff
@@ -7,13 +7,13 @@ class Go
...
-      GOMOD_REGEX = /^(.+)\s+(.+)$/
-      GOMOD_IGNORABLE_REGEX = /^(module\s|require\s+\(|go\s|\))/m
+      GOMOD_REGEX = /^(require\s+)?(.+)\s+(.+)$/
+      GOMOD_IGNORABLE_REGEX = /^(\/\/|module\s|go\s|exclude\s|replace\s|require\s+\(|\))/m
       GOSUM_REGEX = /^(.+)\s+(.+)\s+(.+)$/
 
       def self.mapping
         {
-          // Go Modules (recommended)
+          # Go Modules (recommended)
```
tyrelsouza, line 16: how did this work?!

tyrelsouza: Talked to Tieg, it wasn't deployed yet, makes sense!

Issue #478, opened 2020-08-28
"Include not found" for .csproj if the file has Update reference


sadikay:
If the `.csproj` file like below the analyzer fails and returns `[]` dependencies because of the `Update` reference in the last node.

```
  <ItemGroup>
    <PackageReference Include="Microsoft.AspNetCore" Version="1.1.1" />
    <PackageReference Include="Microsoft.AspNetCore.StaticFiles" Version="2.2.0" />
  </ItemGroup>

  <ItemGroup>
    <PackageReference Update="Microsoft.NETCore.App" Version="1.1.1" />
  </ItemGroup>
```

sadikay:
@katzj any update ?

PR #479, opened 2020-08-28
fix Update reference in .csproj #478

sadikay:
If the .csproj file has `Update` reference the analyzer fails. Because it expect only `Include` reference.

Error message: "Include not found"

Failure case
```
  <ItemGroup>
    <PackageReference Update="Microsoft.NETCore.App" Version="1.1.1" />
  </ItemGroup>
```

### DIFF
--- a/lib/bibliothecary/parsers/nuget.rb
+++ b/lib/bibliothecary/parsers/nuget.rb
@@ -91,7 +91,7 @@ module Bibliothecary
 
       def self.parse_csproj(file_contents)
         manifest = Ox.parse file_contents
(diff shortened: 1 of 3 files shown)


katzj:
What does an `Update` in the context of a PackageReference mean? It's not listed in the [docs](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files)

sadikay:
Looks like a solution for version handling https://www.strathweb.com/2018/07/solution-wide-nuget-package-version-handling-with-msbuild-15/

djpowers (approved):
Realized I was able to resolve the merge conflict directly in the GitHub UI, thanks again!

PR #480, opened 2020-09-08
Ensure original requirement is returned when parsing yarn.lock files

johnbintz-tidelift:
This will help us match up what Yarn versions came from what package.json requirements.


lib/bibliothecary/parsers/npm.rb
```diff
@@ -70,6 +70,7 @@ def self.parse_yarn_lock(file_contents)
           {
             name: dep[:name],
             requirement: dep[:version],
+            original_requirement: dep[:requirement],
```
katzj, line 73: I'm getting a little hung up on the naming here and not loving this as it seems a little confusing. @jsonperl do you have any thoughts/suggestions?

katzj (commented):
I suspect we're going to need this in more parsers (python definitely has some logic where it looks like the same idea would apply) but it's probably reasonable to add as we go 

Definitely feels like at least 6.9.0 worthy

lib/bibliothecary/parsers/npm.rb
kbarrette: I have the feeling we already have a name for this in depci somewhere...

jsonperl: Hmm, so it's basically the requested version in the lockfile. lockfile_requirement?

katzj (approved):

jsonperl (approved):

PR #483, opened 2020-11-24
Parse a Resolved Dependencies json File for Go

mikeyoung85:


lib/bibliothecary/parsers/go.rb
```diff
@@ -140,6 +144,10 @@ def self.parse_go_sum(file_contents)
         deps.uniq
       end
 
+      def self.parse_go_resolved(file_contents)
+        JSON.parse(file_contents).map { |dep| { name: dep["Path"], requirement: dep["Version"], type: 'runtime' } }
```
katzj, line 148: I think we want to skip the ones that are `Main`

mikeyoung85: I'll add a filter in here

lib/bibliothecary/parsers/go.rb
```diff
@@ -140,6 +144,12 @@ def self.parse_go_sum(file_contents)
         deps.uniq
       end
 
+      def self.parse_go_resolved(file_contents)
+        JSON.parse(file_contents)
+          .select { |dep| dep["Main"] == "false" }
```
jsonperl, line 149: Shouldn't matter but maybe `dep["Main"] != "true"` ?

jsonperl (approved):
Wonderful! 🎉

jsonperl (changes requested):
Oh actually, don't forget a Version bump.

PR #491, opened 2021-03-16
Add tree parsing support to

mellisio:


lib/bibliothecary/parsers/maven.rb
```diff
@@ -148,6 +152,16 @@ def self.parse_maven_resolved(file_contents)
           .compact
           .uniq
       end
+      def self.parse_maven_tree(file_contents)
+        captures = file_contents.scan(/\s([\w\.-]+:[\w\.-]+:[\w\.-]+:[\w\.-]+:[a-zA-Z]+)/).flatten.uniq
```
jsonperl, line 156: I'm just gonna go ahead and trust this regex is correct ;)

jsonperl (approved):

lib/bibliothecary/parsers/maven.rb
tiegz: Should be able to replace the `\.`'s with `.` since you don't need to escape in character groups.

tiegz (approved):

PR #496, opened 2021-04-03
Add a tree parser for "npm ls --json" output.

tiegz:


lib/bibliothecary/parsers/npm.rb
```diff
@@ -22,6 +22,11 @@ def self.mapping
           match_filename("package-lock.json") => {
             kind: 'lockfile',
             parser: :parse_package_lock
+          },
+          # The output of `npm ls --json`
+          match_filename("npm-dependency-tree.json") => {
```
tiegz, line 27: I have half a mind to call this file `npm-ls.json` to match `npm-shrinkwrap.json`, but this matches the Maven tree filename.

katzj: I like calling it `npm-ls.json` fwiw

tiegz: Renamed ✅

jsonperl (approved):

PR #498, opened 2021-04-16
Add CircleCI config

tiegz:
I realized a spec from my last PR was broken and it wasn't failing checks. This adds the CircleCI config so we can turn that check on, and it also fixes my broken spec 🌞

### DIFF
--- a/lib/bibliothecary/version.rb
+++ b/lib/bibliothecary/version.rb
@@ -1,3 +1,3 @@
 module Bibliothecary
-  VERSION = "6.10.7"
+  VERSION = "6.10.8"
 end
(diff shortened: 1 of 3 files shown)


lib/bibliothecary/version.rb
```diff
@@ -1,3 +1,3 @@
 module Bibliothecary
-  VERSION = "6.10.7"
+  VERSION = "6.10.8"
```
jsonperl, line 2: Is this a vbump tho?

jsonperl (approved):

lib/bibliothecary/version.rb
tiegz: 🤷🏻‍♂️

PR #499, opened 2021-05-10
Add pip parsing to conda environment yaml files

jsonperl:
Trying out a bit of a hybrid approach where we rely on conda-parser for the primary conda bits, and then added some of the previous code back for parsing pip-in-conda dependencies. I have to dig in a little more here, because I don't think Bibliothecary#analyse_file should be producing an array. 

I'm wondering if we should have the pypi parser do this and figure out some way to let both parsers have at the file.

### DIFF
--- a/lib/bibliothecary.rb
+++ b/lib/bibliothecary.rb
@@ -21,8 +21,8 @@ module Bibliothecary
     runner.load_file_list(path)
   end
 
-  def self.init_package_manager(info)
-    runner.init_package_manager(info)
+  def self.applicable_package_managers(info)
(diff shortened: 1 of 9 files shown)


tyrelsouza:
>I don't think Bibliothecary#analyse_file should be producing an array.

I feel like I agree, I remember we added that to support something, and then ripping out pip support for conda we were able to make it a single object vs an array.

jsonperl:
This allows multiple environments to claim a single file as applicable. Now pypi can process environment.yml files if they happen to contain pypi packages.

PR #505, opened 2021-06-17
maven-dependency-tree.txt: prevent package name from being listed as a dep.

tiegz:
When parsing a `maven-dependency-tree.txt`, the first dep will be the package you're parsing. This skips that first dep so it's not counted as a dep of itself.

### DIFF
--- a/lib/bibliothecary/parsers/maven.rb
+++ b/lib/bibliothecary/parsers/maven.rb
@@ -152,9 +152,12 @@ module Bibliothecary
           .compact
           .uniq
       end
+
       def self.parse_maven_tree(file_contents)
         file_contents = file_contents.gsub(/\r\n?/, "\n")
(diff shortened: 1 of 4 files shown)


lib/bibliothecary/parsers/maven.rb
```diff
@@ -152,9 +152,12 @@ def self.parse_maven_resolved(file_contents)
           .compact
           .uniq
       end
+
       def self.parse_maven_tree(file_contents)
         file_contents = file_contents.gsub(/\r\n?/, "\n")
         captures = file_contents.scan(/^\[INFO\](?:(?:\+-)|\||(?:\\-)|\s)+((?:[\w\.-]+:)+[\w\.\-${}]+)/).flatten.uniq
+        captures.shift if captures.size > 1 # first dep line will be the package itself (unless we're only analyzing a single line)
```
tiegz, line 159: The `size > 1` constraint is because I've seen this method used to parse a single line.

jsonperl (approved):

PR #508, opened 2021-08-30
Add support for Drupal Composer repo by returning Drupal version.

tiegz:
Drupal has a Composer repository at `https://packages.drupal.org/{7,8}`, and most Drupal packages in a Composer manifest will point to that (versus Packagist). 

However, Drupal's repository uses a mapper that translates the semver that Composer asks for into Drupal-compatible versions, e.g. `8-x.2-1` (this example translates to version `2.1.0` that is compatible with Drupal 8).

This change returns the Drupal version instead of the semver version if the package is a Drupal one.

### Example composer.lock snippet:

```
        ...
        {
            "name": "drupal/allowed_formats",
            "version": "1.3.0",
            "source": {
                "type": "git",
                "url": "https://git.drupalcode.org/project/allowed_formats.git",
                "reference": "8.x-1.3"
            },
            ...
         }
         ...
```

After this change, Bibliothecary would return `{name: "drupal/allowed_formats", requirement: "1.3.0", drupal_requirement: "8.x-1.3", type: "runtime"}` from that ^.

https://app.clubhouse.io/tidelift/story/19449/libraries-add-drupal-repository-as-a-source-of-composer-files#activity-19459

### DIFF
--- a/lib/bibliothecary/parsers/packagist.rb
+++ b/lib/bibliothecary/parsers/packagist.rb
@@ -21,15 +21,17 @@ module Bibliothecary
       def self.parse_lockfile(file_contents)
         manifest = JSON.parse file_contents
         manifest.fetch('packages',[]).map do |dependency|
+          requirement = is_drupal_module(dependency) ? dependency.dig("source", "reference") : dependency["version"]
           {
             name: dependency["name"],
-            requirement: dependency["version"],
+            requirement: requirement,
             type: 'runtime'
           }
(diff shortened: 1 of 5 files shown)


tyrelsouza:
I feel like having BOTH makes sense from biblio. can we stiore the   8.x-2.x style and it be picked up? or will we only be ableto store the semver one?

tiegz:
@tyrelsouza I thinkkkk we could return the `source` field from the lockfile too, unless Bibliothecary turns out to be strict about which fields it returns.

PR #510, opened 2021-09-23
Match pip-tools' pip-compile output as a lockfile

tyrelsouza:


katzj (approved):
🔥

katzj:
Just to confirm, we only do the one parse pass and don't fall through to the next if the file happened to be named `requirements.txt`?

tyrelsouza:
That's the way it appears to me from running tests (otherwise it would end up lookin like a manifest?)

PR #513, opened 2021-09-24
[7.1.4] Parse files and contents with load_file_info_list_from_contents

tyrelsouza:


lib/bibliothecary/runner.rb
```diff
@@ -124,5 +134,16 @@ def ignored_dirs
...
+    private
+
+    def add_files_to_list(file_list, info)
+      applicable_package_managers(info).each do |package_manager|
+        file = info.dup
+        file.package_manager = package_manager
+
+        file_list.push(file)
+      end
+    end
```
tyrelsouza, line 147: this seems to work to refactor because ruby is pass by ref not pass by value?

lib/bibliothecary/runner.rb
```diff
@@ -42,6 +42,22 @@ def package_managers
       Bibliothecary::Parsers.constants.map{|c| Bibliothecary::Parsers.const_get(c) }.sort_by{|c| c.to_s.downcase }
     end
 
+    def load_file_info_list_from_files(files)
+      # Parses an array of format [{file_path: "", contents: ""},] to match
+      #  on both filename matches, and one content_match patterns.
+      file_list = []
+
+      files.each do |file|
+        info = FileInfo.new(nil, file[:file_path], file[:contents])
```
tyrelsouza, line 51: This is the MAIN change here, adding contents to come in here if we have the contents.

in the FileInfo stuff its read from a temp location, so setting here does all the checks properly.

tyrelsouza: everythgin else is just new methods to explicitly call this out
(vs just checking "is array of strings, vs is array of objects" shenanigans)

lib/bibliothecary/runner.rb
```diff
@@ -42,6 +42,22 @@ def package_managers
       Bibliothecary::Parsers.constants.map{|c| Bibliothecary::Parsers.const_get(c) }.sort_by{|c| c.to_s.downcase }
     end
 
+    def load_file_info_list_from_files(files)
```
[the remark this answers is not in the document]
tyrelsouza: Ah, where was _I_ wouldn't think files _wouldn't_. If we're leaving this to semantics, then I will defer to you and Tieg.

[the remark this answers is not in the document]
tiegz: Yeah, I actually had the same thought when I was looking at this earlier. I kinda like `load_file_info_list_with_contents` more, but `load_file_info_list_from_contents` makes sense bc it's consistent with the other method `load_file_info_list_from_paths`.

Thanks for including the docs below, btw!

tiegz (approved):

katzj (approved):

PR #514, opened 2021-09-27
PyPI - Parse the versions from requirement[extra]==version format 

tyrelsouza:


lib/bibliothecary/parsers/pypi.rb
```diff
@@ -4,7 +4,12 @@ class Pypi
       include Bibliothecary::Analyser
 
       INSTALL_REGEXP = /install_requires\s*=\s*\[([\s\S]*?)\]/
-      REQUIRE_REGEXP = /([a-zA-Z0-9]+[a-zA-Z0-9\-_\.]+)([><=\w\.,]+)?/
+
+      # Capture Group 1 is package.
+      # Optional Group 2 is [extras].
+      # Capture Group 3 is Version
+      REQUIRE_REGEXP = /([a-zA-Z0-9]+[a-zA-Z0-9\-_\.]+)(?:\[.*?\])*([><=\w\.,]+)?/
```
tyrelsouza, line 11: `(?:\[.*?\])*` is new between the match groups, as match group 2 that we ignore.

katzj (approved):

lib/bibliothecary/parsers/pypi.rb
```diff
@@ -160,7 +165,7 @@ def self.parse_setup_py(manifest)
           next unless match
           deps << {
             name: match[1],
-            requirement: match[2] || '*',
+            requirement: match[-1] || '*',
```
tyrelsouza, line 168: Sometimes we're len2 sometimes we're len3, so take last item.

PR #526, opened 2022-04-17
Adds support for gradle-kotlin-dependencies-q.txt

tiegz:
The new `gradle-kotlin-dependencies-q.txt` file will be the output of `gradle dependencies -q` on a `build.gradle.kts` file. 

It should be the same as the output from `build.gradle` AFAIK, but this uses a different name just in case there *are* differences, and also to denote that it came from a `build.gradle.kts` file.

(support for parsing `build.gradle.kts` is a bigger task and will come in a later PR)

### DIFF
--- a/spec/fixtures/gradle-kotlin-dependencies-q.txt
+++ b/spec/fixtures/gradle-kotlin-dependencies-q.txt
@@ -1,0 +1,312 @@
+
+------------------------------------------------------------
+Project ':app'
+------------------------------------------------------------
+
+annotationProcessor - Annotation processors and their dependencies for source set 'main'.
+No dependencies
+
+api - API dependencies for compilation 'main' (target  (jvm)). (n)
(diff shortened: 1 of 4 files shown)


spec/fixtures/gradle-kotlin-dependencies-q.txt
```diff
@@ -0,0 +1,312 @@
+
```
tiegz, line 1: FTR, this file was generated via `gradle init`, and selecting a Kotlin application with the Kotlin gradle DSL.

tiegz:
I'm rethinking this now -- maybe it's fine to just use `gradle-dependencies-q.txt` for both Groovy and Kotlin output. Anyone have thoughts?

tiegz:
Yeah, I should probably use the same file name. Going to open up another PR

PR #529, opened 2022-04-22
[8.0.0] Add CycloneDX support

johnbintz-tidelift:
I ran the new code against these sets of SBOMs and everything appeared to parse correctly, or at least parsed without error:

* https://github.com/CycloneDX/bom-examples/tree/master/SBOM
* https://github.com/sonatype-nexus-community/cyclonedx-sbom-examples


havocp:
I remember the "multiple platforms in a file" problem having some significant complexity (the dependencies.csv parser sort of punts on it for some reason).

I can't look at this in detail for an hour or two but would like to understand that aspect.

There was a more comprehensive support for multiple platforms that was added and then reverted historically.

johnbintz-tidelift:
@havocp I reworked the approach I was using to make it so you include (well, extend) a Parser with the CycloneDX parser and then all the CycloneDX bits are added to the individual parsers. This means that, right now, when parsing a CycloneDX file, there's a bunch of matching plartforms that are totally blank.

lib/bibliothecary/analyser.rb
```diff
@@ -24,6 +24,7 @@ def self.create_analysis(platform_name, relative_path, kind, dependencies)
     def self.included(base)
       base.extend(ClassMethods)
     end
+
     module ClassMethods
       def generic?
         platform_name == "generic"
```
havocp, line 30: I'm not sure when this "generic?" predicate is used if at all, but I wonder whether when it's used, someone would be expecting it to also match cyclonedx. It's possible I added it thinking it would enable special-casing a multi-platform dependencies.csv and then ended up not succeeding in using this.

lib/bibliothecary/multi_parsers/cyclonedx.rb
```diff
@@ -0,0 +1,131 @@
...
+          purl = PackageURL.parse(component["purl"])
+
+          obj << purl
+        end
+
+        entries[platform_name.to_sym]
+      end
+
+      def parse_cyclonedx_xml(file_contents)
+        manifest = Ox.parse(file_contents)
```
havocp, line 106: we're going to end up parsing the json or xml a dozen times I think? (once per platform?)
Any ideas how we could avoid that?

I think the approach here seems reasonable except for that performance hit seems sort of bad.

lib/bibliothecary/analyser.rb
havocp: so one option could be (maybe in its own PR) make dependencies.csv work the same as cyclonedx does now and drop this "generic " idea.

johnbintz-tidelift:
@havocp I rearranged a bunch of furniture to help me understand the data flows, renamed a bunch of things for improved consistency, and implemented caching within `Bibliothecary::Runner`.

spec/parsers/maven_spec.rb
```diff
@@ -308,23 +310,27 @@
     expect(described_class.match?(fixture_path('ivy_reports/invalid_syntax.xml'))).to be_falsey
   end
 
-  it 'uses parent properties during resolve' do
-    parent_props = {"bibliothecary.version"=>"9.9.9"}
-    deps = described_class.parse_pom_manifest(load_fixture('pom.xml'), parent_props)
+  xdescribe 'parent properties' do
+    # TODO: This parent properties feature doesn't seem to be used anywhere
+    # in the codebase. Is it needed?
```
havocp, line 315: I wonder if this spec fails if you set `parent_props={}`, that could clarify whether it's actually doing anything?

lib/bibliothecary/multi_parsers/cyclonedx.rb
```diff
@@ -85,7 +85,11 @@ def self.mapping
       end
 
       def parse_cyclonedx_json(file_contents, options: {})
-        manifest = JSON.parse(file_contents)
+        manifest = nil
+
+        manifest = try_cache(options, :cyclonedx_json) do
```
havocp, line 91: is `options` shared among multiple files? Should the cache key have the path in it?

havocp (commented):
Nice to see some cleanup here! I'd want to be sure we can parse multiple cyclonedx files in a single file tree; that's the only substantive comment I think.
Like foo/cyclonedx.xml and bar/cyclonedx.xml both passed in at once.

lib/bibliothecary/multi_parsers/ruby_manifest.rb
```diff
@@ -0,0 +1,20 @@
+module Bibliothecary
+  module MultiParsers
+    module RubyManifest
```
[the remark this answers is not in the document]
johnbintz-tidelift: Done and I renamed the file as well.

havocp (approved):

havocp (approved):

johnbintz-tidelift:
I see what's going on with the pom stuff. There's a parsing method that's doing double duty (acting as a target for `parse_file` and also being called by something else) and the something else is adding the parent properties. I'll get that cleaned up.

johnbintz-tidelift:
OK I found this: https://github.com/librariesio/libraries.io/blob/e970925aade2596a03268b6e1be785eba8502c62/app/models/package_manager/maven.rb#L129

A direct call to parse_pom_manifest in Libraries. I'll shuffle some things around to make sure Libraries won't break when I update this.

lib/bibliothecary/parsers/maven.rb
```diff
@@ -197,10 +197,13 @@ def self.parse_resolved_dep_line(line)
         }
       end
 
-      def self.parse_pom_manifest(file_contents, options: {})
-        # TODO: is this even used?
-        parent_properties = {}
+      def self.parse_standalone_pom_manifest(file_contents, options: ())
```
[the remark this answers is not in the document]
johnbintz-tidelift: No that is a typo.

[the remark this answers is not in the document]
havocp: ruby does seem to accept it as nil, maybe some kind of nod to lisp :-)

havocp (approved):

PR #532, opened 2022-04-29
[8.2.0] Implements a super-simple build.gradle.kts parser.

tiegz:
Adds a super simplified parser for `build.gradle.kts` files using regexes.

Deps like this **will be caught**:

```
dependencies { 
    implementation("com.google.guava:guava")
    implementation("com.google.guava:guava:30.1.1-jre")
}
```

Deps like these **will not be caught**:

```
dependencies {
      "implementation"("org.jetbrains.kotlin:kotlin-bom")
      implementation(project(SomeClass.SOME_CONSTANT))
      implementation(platform("org.jetbrains.kotlin:kotlin-bom"))
      implementation(group = "org.jetbrains.kotlin", name = "kotlin-bom", version = "1.2.3")
      implementation(npm("text-encoding", "0.7.0"))
      "js"("jquery:jquery:3.2.1@js")
}
```


spec/fixtures/build.gradle.kts
```diff
@@ -1,122 +1,42 @@
 /*
- * Copyright 2014 Google Inc.
```
tiegz, line 2: This was a `build.gradle` file that I basically added parentheses to, but then realized that kotlin's differences are much more nuanced, so I swapped the file with a basic boilerplate `build.gradle.kts`.

johnbintz-tidelift:
Not related to this PR but to manifest parsing in general: what approach do we want to take when we get more complex in our parsing? Something like [EBNF](https://github.com/dryruby/ebnf)? (And if we get lucky, could we find existing grammars to use for these)

lib/bibliothecary/parsers/maven.rb
```diff
@@ -236,8 +247,17 @@ def self.parse_gradle(file_contents, options: {})
...
-        []
+        file_contents
+          .scan(GRADLE_KTS_SIMPLE_REGEX)                                                   # match 'implementation("group:artifactId:version")'
+          .map { |(_type, dep_match)| GRADLE_KTS_GAV_REGEX.match(dep_match) }              # extract ["group", "artifactId", ?"version"]
+          .reject { |gav_match| gav_match.nil? || gav_match[1].nil? || gav_match[2].nil? } # remove any with missing group/artifactId
+          .map { |gav_match|
+            {
+              name: [gav_match[1], gav_match[2]].join(":"),
+              requirement: gav_match[3] || "*",
+              type: nil # TODO: we may be able to infer dep tyeps using the _type var above.
```
johnbintz-tidelift, line 258:
```suggestion
              type: nil # TODO: we may be able to infer dep types using the _type var above.
```

tiegz:
(I could also add some of the not-supported cases I added above into the regexp if anyone thinks it's worth it, but I just didn't need them for my use case right now)

johnbintz-tidelift (approved):

PR #534, opened 2022-05-04
[8.2.1] Expand simple parsing support for build.gradle.kts

tiegz:
* refactor the `build.gradle.kts` parsing regexes 
* add support for interpolated verisons, e.g. `group:artifactId:${someVar["my-version"]}`
* and start returning the method used for the dep in`type` , e.g. "implementation". This is what the `build.gradle` parser currently does as well

Folllowup to https://github.com/librariesio/bibliothecary/pull/532

### DIFF
--- a/spec/fixtures/build.gradle.kts
+++ b/spec/fixtures/build.gradle.kts
@@ -34,6 +34,11 @@ dependencies {
 
     // Use the Kotlin JUnit integration.
     testImplementation(     "org.jetbrains.kotlin:kotlin-test-junit:1.0.0"     )
+
+    implementation(project(":my-internal-lib"))
+
+    // --- Androidx ---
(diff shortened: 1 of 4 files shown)


spec/fixtures/build.gradle.kts
```diff
@@ -34,6 +34,11 @@ dependencies {
 
     // Use the Kotlin JUnit integration.
     testImplementation(     "org.jetbrains.kotlin:kotlin-test-junit:1.0.0"     )
+
+    implementation(project(":my-internal-lib"))
```
tiegz, line 38: May add support for these later, depending on whether we want to consider projects as "packages" (i.e. they show up as a node in the dep tree From `gradle dependencies -q`)

johnbintz-tidelift (approved):

PR #537, opened 2022-05-06
Support VCS URLs in PyPI requirements.txt files

johnbintz-tidelift:


lib/bibliothecary/parsers/pypi.rb
```diff
@@ -179,20 +179,51 @@ def self.parse_setup_py(file_contents, options: {})
         deps
       end
 
+      NoEggSpecified = Class.new(ArgumentError)
```
papiro, line 182: What is egg?

papiro (approved):

lib/bibliothecary/parsers/pypi.rb
johnbintz-tidelift: An "egg" is the name of an older format for distributing Python code, nowadays mostly replaced by "wheel"s. However, PyPI uses `egg` in the fragment of the URL to identify which package in the PyPI index the git repo represents.

https://stackoverflow.com/a/55526482

papiro (approved):

PR #538, opened 2022-05-13
Don't group multimanifest files with normal manifests

johnbintz-tidelift:
Story: https://app.shortcut.com/tidelift/story/24036/if-biblio-doesn-t-return-the-expected-groups-then-i-think-the-fix-would-be-to-go-make-a-failing-test


papiro (approved):
LGTM

PR #549, opened 2022-07-11
Update Cargo parser to handle dev dependencies

djpowers:
This ensures that Cargo development dependencies are correctly parsed, which are ignored at present. (Probably easiest to review commit by commit.)

See [The Cargo Book: Specifying Dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#development-dependencies) for more info.


tiegz (approved):
Bummer that the lockfile doesn't denote this!

PR #551, opened 2022-07-12
Update NuGet parser to handle dev dependencies

djpowers:
This adds support for the `developmentDependency` attribute from `packages.config` files:
- https://docs.microsoft.com/en-us/nuget/reference/packages-config#examples

It also adds support `PackageReference`elements as used in `.csproj` files:
- https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets

(Edit)
It also adds `developmentDependency` support for use in `.nuspec` files:
- https://docs.microsoft.com/en-us/nuget/reference/nuspec#developmentdependency

### DIFF
--- a/lib/bibliothecary/parsers/nuget.rb
+++ b/lib/bibliothecary/parsers/nuget.rb
@@ -108,7 +108,7 @@ module Bibliothecary
             requirement = dependency.nodes.detect{ |n| n.value == "Version" }&.text
           end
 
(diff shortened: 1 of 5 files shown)


tiegz (commented):
Looks good! Any reason to not parse them out of .nuspec files too?

tiegz (approved):

PR #555, opened 2022-07-18
NPM: omit "//" comment keys in package.json

tiegz:
I stumbled across an obscure link to a mailing list that comments like this in `package.json` are valid, so we can skip them inside `dependencies`/etc when parsing deps.

### DIFF
--- a/lib/bibliothecary/parsers/npm.rb
+++ b/lib/bibliothecary/parsers/npm.rb
@@ -74,8 +74,12 @@ module Bibliothecary
       def self.parse_manifest(file_contents, options: {})
         manifest = JSON.parse(file_contents)
         raise "appears to be a lockfile rather than manifest format" if manifest.key?('lockfileVersion')
-        map_dependencies(manifest, 'dependencies', 'runtime') +
-        map_dependencies(manifest, 'devDependencies', 'development')
+        
+        (
+          map_dependencies(manifest, 'dependencies', 'runtime') +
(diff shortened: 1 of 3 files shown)


lib/bibliothecary/parsers/npm.rb
```diff
@@ -74,8 +74,12 @@ def self.parse_package_lock_deps_recursively(dependencies, depth=1)
...
         manifest = JSON.parse(file_contents)
         raise "appears to be a lockfile rather than manifest format" if manifest.key?('lockfileVersion')
-        map_dependencies(manifest, 'dependencies', 'runtime') +
-        map_dependencies(manifest, 'devDependencies', 'development')
+        
+        (
+          map_dependencies(manifest, 'dependencies', 'runtime') +
+          map_dependencies(manifest, 'devDependencies', 'development')
+        )
+          .reject { |dep| dep[:name] == "//" } # Omit comment keys. They are valid in package.json: https://groups.google.com/g/nodejs/c/NmL7jdeuw0M/m/yTqI05DRQrIJ
```
TravisTX, line 82: Is npm open source? Might be handy to find the source for this, and confirm there's no other special cases for it. Like whitespace on either side of the "//" for example.

tiegz: I tried to, and can't find anything 😆 I couldn't find anything in node either, but I'm going to check again

PR #547, opened 2022-07-10
Capture original_{name,requirement} in maven parser, for gradle aliases.

tiegz:
When we see an alias in `gradle-dependencies-q.txt`, capture `:original_name` and `:original_requirement`. For example:

`|    +--- apache:commons-io:1.4 -> commons-io:commons-io:2.6`

``` ruby
{
  original_name: "apache:commons-io",
  original_requirement: "1.4",
  name: "commons-io:commons-io",
  requirement: "2.6",
  type: "runtimeClassPath",  
}
```

NB Bibliothecary already uses the `:original_name` key for transforming Packagist/Drupal package names.

PR #557, opened 2022-08-04
Handle gradle renames that don't have a requirement in manifest.

tiegz:
Followup to https://github.com/librariesio/bibliothecary/pull/544 and https://github.com/librariesio/bibliothecary/pull/547

Sometimes a rename in `gradle-dependencies-q.txt` won't have a version requirement in the original manifest, so this adds support to recognize those cases as well.

### DIFF
--- a/spec/parsers/maven_spec.rb
+++ b/spec/parsers/maven_spec.rb
@@ -422,19 +422,20 @@ RSpec.describe Bibliothecary::Parsers::Maven do
 
       runtime_classpath = deps[:dependencies].select {|item| item[:type] == "runtimeClasspath"}
 
-      expect(runtime_classpath.length).to eq 156
+      expect(runtime_classpath.length).to eq 157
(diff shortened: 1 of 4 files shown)


spec/parsers/maven_spec.rb
```diff
@@ -422,19 +422,20 @@
...
-      renamed_dep = runtime_classpath.
-        select do |item| 
-          item[:name] == "commons-io:commons-io" && 
-          item[:original_name] == "apache:commons-io" &&
-          item[:requirement] == "2.6" &&
-          item[:original_requirement] == "1.4" 
+      # test rename resolutions
+      [
+        {name: "commons-io:commons-io", requirement: "2.6", original_name: "apache:commons-io", original_requirement: "1.4"},
+        {name: "axis:axis", requirement: "1.4", original_name: "apache:axis", original_requirement: "*"}
```
tiegz, line 431: We also default a nil requirement in `build.gradle` files to `*` so I'm using `*` here too, instead of nil.

katzj (approved):

PR #562, opened 2022-08-21
Gradle lockfile parser: extract the line-ending regexp, and remove interpolation logic.

tiegz:
* extract line-ending regexp so it's reusable elsewhere
* also, remove the "interpolate the project name" logic that was added in https://github.com/librariesio/bibliothecary/pull/560, because these are cycles [as confirmed by the gradle tests](https://github.com/gradle/gradle/blob/d1afb35abdbb2065de3086da885df8071b30049a/subprojects/diagnostics/src/integTest/groovy/org/gradle/api/tasks/diagnostics/DependencyInsightReportTaskIntegrationTest.groovy#L1697-L1753), so there's no reason to track them

### DIFF
--- a/lib/bibliothecary/parsers/maven.rb
+++ b/lib/bibliothecary/parsers/maven.rb
@@ -14,7 +14,7 @@ module Bibliothecary
 
       # Project declaration lines so we know the current project name
       # e.g. "Project ':submodules:test'" (this example is a project nested in submodules/test/ folder)
-      GRADLE_PROJECT_DECLARATION_REGEX = /(?:Root project|Project) '?:([^\s']+)'?/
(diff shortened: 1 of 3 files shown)


lib/bibliothecary/parsers/maven.rb
```diff
@@ -12,15 +12,15 @@ class Maven
       # e.g. "|    \\--- com.google.guava:guava:23.5-jre (*)"
       GRADLE_DEP_REGEX = /(\+---|\\---){1}/
 
-      # Project declaration lines so we know the current project name
-      # e.g. "Project ':submodules:test'" (this example is a project nested in submodules/test/ folder)
-      GRADLE_PROJECT_DECLARATION_REGEX = /Project '?:([^\s']+)'?/
```
tiegz, line 17: Since we don't interpolate the project name anymore, there's no need to track it.

mikeyoung85 (approved):

PR #564, opened 2022-11-07
Remove Byte Order Marks from beginning of file contents before parsing.

tiegz:
Fixes parsing of files like this cyclonedx.json, which have a Byte Order Mark at the beginning of the file:

``` ruby
File.read("cyclonedx.json").unpack("U*")[0,10]
=> [65279, 123, 13, 10, 32, 32, 32, 32, 34, 98]

JSON.parse(File.read("cyclonedx.json"))
JSON::ParserError: Empty input (after ) at line 1, column 1 [parse.c:1116] in '{
    "bomFormat": "CycloneDX",
    "specVersion": "1.4",
    "version": 1,
    "metadata": {
        "tools": [
            {
                "vendor": "CycloneDX",
                "name": "Node.js module",
```

### DIFF
--- a/lib/bibliothecary/file_info.rb
+++ b/lib/bibliothecary/file_info.rb
@@ -17,7 +17,8 @@ module Bibliothecary
             # file that's actually on the filesystem
             nil
           else
-            File.open(@full_path).read
+            # Remove any Byte Order Marks so JSON, etc don't fail while reading them.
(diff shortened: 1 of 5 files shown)


tiegz:
> so could we say... we are removing the BOM so we can parse this BOM?

💯

PR #566, opened 2022-11-23
Adds support for NPM lockfileVersion 3

tiegz:
`lockfileVersion: "3"` seems to be mandatory in npm v9+, and it drops support for the `"dependencies"` object in favor of the `"packages"` object:

https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json#lockfileversion


djpowers (approved):
LGTM, and the comments are much appreciated.

PR #568, opened 2023-02-03
pyproject.toml: support PEP621 projects too

tiegz:
`pyproject.toml` can include poetry metadata, but it can also include project metadata according to [PEP 621](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/), under the `[project]` field. This adds support for those `[project.dependencies]` deps.

### DIFF
--- a/lib/bibliothecary/parsers/pypi.rb
+++ b/lib/bibliothecary/parsers/pypi.rb
@@ -108,6 +108,8 @@ module Bibliothecary
         pep621_deps = pep621_manifest.fetch('dependencies', []).map { |d| parse_pep_508_dep_spec(d) }
         deps += map_dependencies(pep621_deps, 'runtime')
 
+        # We're combining both poetry+PEP621 deps instead of making them mutually exclusive, until we
+        # find a reason not to ingest them both.
         deps
       end
 
--- a/lib/bibliothecary/parsers/pypi.rb
+++ b/lib/bibliothecary/parsers/pypi.rb
@@ -14,6 +14,9 @@ module Bibliothecary
       MANIFEST_REGEXP = /.*require[^\/]*(\/)?[^\/]*\.(txt|pip|in)$/
       PIP_COMPILE_REGEXP = /.*require.*$/
 
+      # Adapted from https://peps.python.org/pep-0508/#names
+      PEP_508_NAME_REGEX = /^([A-Z0-9][A-Z0-9._-]*[A-Z0-9]|[A-Z0-9])/i
+
       def self.mapping
         {
           match_filenames('requirements-dev.txt', 'requirements/dev.txt',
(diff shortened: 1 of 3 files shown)


spec/parsers/pypi_spec.rb
```diff
@@ -309,6 +309,37 @@
...
+]    
+    FILE
+
+    expect(described_class.analyse_contents('pyproject.toml', source)).to eq({
+      platform: "pypi",
+      path: "pyproject.toml",
+      dependencies: [
+        {name: "black", requirement: "*", type: "runtime"}, 
+        {name: "isort", requirement: "*", type: "runtime"}, 
+        {name: "click == 8.1.3", requirement: "*", type: "runtime"},
```
tiegz, line 334: woops, gotta fix this first

lib/bibliothecary/parsers/pypi.rb
```diff
@@ -14,6 +14,9 @@ class Pypi
       MANIFEST_REGEXP = /.*require[^\/]*(\/)?[^\/]*\.(txt|pip|in)$/
       PIP_COMPILE_REGEXP = /.*require.*$/
 
+      # Adapted from https://peps.python.org/pep-0508/#names
+      PEP_508_NAME_REGEX = /^([A-Z0-9][A-Z0-9._-]*[A-Z0-9]|[A-Z0-9])/i
```
tiegz, line 18: cc @papiro @mpace965

lib/bibliothecary/parsers/pypi.rb
```diff
@@ -90,9 +93,28 @@ def self.parse_pipfile(file_contents, options: {})
...
+        
+        # Parse poetry [tool.poetry] deps
+        poetry_manifest = file_contents.fetch('tool', {}).fetch('poetry', {})
+        deps += map_dependencies(poetry_manifest['dependencies'], 'runtime')
+        deps += map_dependencies(poetry_manifest['dev-dependencies'], 'develop')
+
+        # Parse PEP621 [project] deps
+        pep621_manifest = file_contents.fetch('project', {})
+        pep621_deps = pep621_manifest.fetch('dependencies', []).map { |d| parse_pep_508_dep_spec(d) }
+        deps += map_dependencies(pep621_deps, 'runtime')
```
mpace965, line 109: should these be mutually excluded or is it possible for a single manifest to specify both of these dependencies?

papiro (approved):

lib/bibliothecary/parsers/pypi.rb
tiegz: PEP621 is the "tool-agnostic" way to declare build system requirements in pyproject.toml, and from what I understand Poetry is the only build system that doesn't support it yet. They're discussing it tho, and it looks like one of the examples includes both the "tool.poetry" And "project" namespaces for declaring deps: https://github.com/python-poetry/roadmap/issues/3

tiegz: ah yeah, just found an example here. The deps are the same in both section, but still seems safer to combine them both until we have a reason not to: https://sourcegraph.com/github.com/IceTiki/ruoli-sign-optimization/-/blob/pyproject.toml?subtree=true

mpace965: In that case, does dep need to be deduped? I'm not sure what clients of this method are expecting

tiegz: ah yes! Good call, I think we do that higher up in `Bibliothecary.analyse_file` but it does seem like some of the platform-specific parsers will uniq too.

PR #572, opened 2023-05-26
Add support for optional field in Maven POM parser

kjmahalingam:
In coordination with [this documentation](https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html#how-do-i-use-the-optional-tag), this is to add support for the optional `optional` field in the Maven POM dependency parser.

### DIFF
--- a/lib/bibliothecary/parsers/maven.rb
+++ b/lib/bibliothecary/parsers/maven.rb
@@ -268,11 +268,15 @@ module Bibliothecary
         [].tap do |deps|
           ['dependencies/dependency', 'dependencyManagement/dependencies/dependency'].each do |deps_xpath|
             xml.locate(deps_xpath).each do |dep|
-              deps.push({
+              dep_hash = {
                 name: "#{extract_pom_dep_info(xml, dep, 'groupId', parent_properties)}:#{extract_pom_dep_info(xml, dep, 'artifactId', parent_properties)}",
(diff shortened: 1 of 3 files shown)


lib/bibliothecary/parsers/maven.rb
```diff
@@ -268,11 +268,15 @@ def self.parse_pom_manifest(file_contents, parent_properties = {}, options: {})
...
+              dep_hash = {
                 name: "#{extract_pom_dep_info(xml, dep, 'groupId', parent_properties)}:#{extract_pom_dep_info(xml, dep, 'artifactId', parent_properties)}",
                 requirement: extract_pom_dep_info(xml, dep, 'version', parent_properties),
-                type: extract_pom_dep_info(xml, dep, 'scope', parent_properties) || 'runtime'
-              })
+                type: extract_pom_dep_info(xml, dep, 'scope', parent_properties) || 'runtime',
+              }
+              # optional field is, itself, optional, and will be either "true" or "false"
+              optional = extract_pom_dep_info(xml, dep, 'optional', parent_properties)
+              dep_hash[:optional] = optional == "true" unless optional.nil?
```
kjmahalingam, line 278: We *could* just always include the `optional` field, and set it to `false` by default, but it seems unnecessary unless there's a reason someone can think of to do so.

tiegz (approved):

djpowers (approved):

macowie (approved):

PR #578, opened 2023-09-09
Tz/fix dependency management section in maven

tiegz:
In version 6.6.0 we added support for the `<dependencyManagement>` section in https://github.com/librariesio/bibliothecary/pull/451 . This section is only for dependency configuration though, not for explicitly importing deps into the project. 

This change reverts that behavior so we no longer import those deps, but use the information (version/scope) in there as fallbacks for the actual deps in `<dependencies>` section instead.

https://github.com/librariesio/bibliothecary/pull/451


spec/fixtures/pom.xml
```diff
@@ -90,6 +90,12 @@
                 <artifactId>commons-lang</artifactId>
                 <version>2.6</version>
             </dependency>
+
+            <dependency>
+                <groupId>com.google.guava</groupId>
+                <artifactId>guava</artifactId>
+                <version>${guava.version}</version>
+            </dependency>
```
tiegz, line 98: here I'm moving the `<version>` field for guava from `<dependencies>` into `<dependencyManagement>` to assert that we'll use the latter as fallback.

spec/parsers/maven_spec.rb
```diff
@@ -93,9 +94,6 @@
         { name: "io.libraries:recursive", requirement: "${recursive.test}", type: "runtime" },
         { name: "io.libraries:optional", requirement: "${optional.test}", type: "runtime", optional: true },
         { name: "io.libraries:not-optional", requirement: "${not-optional.test}", type: "runtime", optional: false },
-        # From dependencyManagement section
-        { name: "org.apache.ant:ant", requirement: "1.9.2", type: "runtime" },
-        { name: "commons-lang:commons-lang",requirement: "2.6", type: "runtime" }
```
tiegz, line 98: these were only found in the `<dependencyManagement>` section but not in `<dependencies>`, so Maven wouldn't add these to the project

lib/bibliothecary/version.rb
```diff
@@ -1,3 +1,3 @@
 module Bibliothecary
-  VERSION = "8.6.5"
+  VERSION = "8.7.0"
```
tiegz, line 2: this change doesn't break compatibility, but it could mean fewer dependencies from maven projects, so I'm considering this a bugfix / change in behavior, hence the minor version bump.

mikeyoung85 (approved):

PR #584, opened 2023-12-09
8.7.4: Add support for Poetry 1.2.0+'s dependency groups

tiegz:
Poetry 1.2.0 added support for dependency groups in https://github.com/python-poetry/poetry/pull/4260 , e.g.:

```
[tool.poetry.group.test.dependencies]
pytest = "^6.0.0"
pytest-mock = "*"
```

docs: https://github.com/python-poetry/poetry/blob/1.7.1/docs/managing-dependencies.md#dependency-groups

### DIFF
--- a/lib/bibliothecary/parsers/pypi.rb
+++ b/lib/bibliothecary/parsers/pypi.rb
@@ -101,8 +101,16 @@ module Bibliothecary
 
         # Parse poetry [tool.poetry] deps
         poetry_manifest = file_contents.fetch('tool', {}).fetch('poetry', {})
+        # Poetry 1.0.0-1.2.0 way of defining dev deps
         deps += map_dependencies(poetry_manifest['dependencies'], 'runtime')
-        deps += map_dependencies(poetry_manifest['dev-dependencies'], 'develop')
+        deps += map_dependencies(poetry_manifest['dev-dependencies'], 'development')
+        # Poetry's 1.2.0+ of defining dev deps
+        poetry_manifest
+          .fetch("group", {})
+          .each_pair do |group_name, obj|
+            group_name = "development" if group_name == "dev"
+            deps += map_dependencies(obj.fetch("dependencies", {}), group_name)
+          end
 
         # Parse PEP621 [project] deps
         pep621_manifest = file_contents.fetch('project', {})
--- a/lib/bibliothecary/parsers/pypi.rb
+++ b/lib/bibliothecary/parsers/pypi.rb
@@ -101,14 +101,14 @@ module Bibliothecary
 
         # Parse poetry [tool.poetry] deps
         poetry_manifest = file_contents.fetch('tool', {}).fetch('poetry', {})
-        # Poetry 1.0.0-1.2.0 way of defining dev deps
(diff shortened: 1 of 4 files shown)


lib/bibliothecary/parsers/pypi.rb
```diff
@@ -179,10 +187,10 @@ def self.parse_poetry_lock(file_contents, options: {})
         manifest["package"].each do |package|
           # next if group == "_meta"
           group = case package['category']
-                  when 'main'
-                    'runtime'
                   when 'dev'
                     'develop'
+                  else
+                    'runtime'
```
tiegz, line 193: this ensures that `pypoetry.lock` should not get a `type: nil` back

spec/fixtures/pyproject.toml
```diff
@@ -8,9 +8,14 @@ authors = ["Tyrel Souza <tyrel@tidelift.com>"]
 python = "^3.7"
 django = "^3.0.7"
 
+# Old way to define dev deps (<1.2.0)
 [tool.poetry.dev-dependencies]
 pytest = "^5.2"
 
+# New way to define dev deps (>=1.2.0)
+[tool.poetry.group.dev.dependencies]
```
mikeyoung85, line 16: We should add another group here and test that the type works correctly. Something like `[tool.poetry.group.test.dependencies]`.

lib/bibliothecary/parsers/pypi.rb
```diff
@@ -102,7 +102,15 @@ def self.parse_pyproject(file_contents, options: {})
         # Parse poetry [tool.poetry] deps
         poetry_manifest = file_contents.fetch('tool', {}).fetch('poetry', {})
         deps += map_dependencies(poetry_manifest['dependencies'], 'runtime')
+        # Poetry 1.0.0-1.2.0 way of defining dev deps
         deps += map_dependencies(poetry_manifest['dev-dependencies'], 'develop')
+        # Poetry's 1.2.0+ of defining dev deps
+        poetry_manifest
+          .fetch("group", {})
+          .each_pair do |group_name, obj|
+            group_name = "develop" if group_name == "dev"
```
mikeyoung85, line 111: I'm familiar with toml and how these are getting parsed. What would the expected group names look like. From the Poetry example what would the `group_name` for `[tool.poetry.group.test.dependencies]` resolve to here?

tiegz: AFAICT it'll use the name that was parsed, so `test` would be the group there. The docs say:

> use a tool.poetry.group.<group> section where <group> is the name of your dependency group (for instance, test):

There's an implicit `main` group that is the default, but we've apparently mapped that to `runtime` until now so I'm setting that as the default in the change below.

Also noteworthy is that anything that's not `main` is considered development semantically, so there's no real distinction between `dev` and `test` to Poetry other than the names:

>  Dependency groups, other than the implicit main group, must only contain dependencies you need in your development process. Installing them is only possible by using Poetry.

And FWIW, I'd prefer to use "development" here to be consistent with the other parsers, but there are 3 uses of "develop" in this file so I'm hesitant to break anything for consumers of the gem : \

mikeyoung85 (approved):

PR #585, opened 2024-01-04
8.7.5: go.mod: adds official support for "replace" directive

tiegz:
this should fix the `replace` parsing bugs, where we currently get `{name: "foo -> bar", ...}` back for deps:

* adds parsing support for all go.mod directives: `require`, `replace`, `extract`, and `retract`
* uses `replace` results to replace any matching deps in `require` results
* ignore `exclude` results for now, since they're not related to `require` results

### DIFF
--- a/lib/bibliothecary/parsers/go.rb
+++ b/lib/bibliothecary/parsers/go.rb
@@ -13,6 +13,7 @@ module Bibliothecary
       GOMOD_MULTILINE_DEP_REGEXP = /^#{GOMOD_DEP_REGEXP}.*$/
       GOMOD_MULTILINE_START_REGEXP = /^(?<category>require|exclude|replace|retract)\s+\(/
       GOMOD_MULTILINE_END_REGEXP = /^\)/
+      GOMOD_COMMENT_REGEXP = /(\/\/(.*))/
       GOSUM_REGEXP = /^(.+)\s+(.+)\s+(.+)$/
 
       def self.mapping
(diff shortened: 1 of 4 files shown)


lib/bibliothecary/parsers/go.rb
```diff
@@ -7,8 +7,12 @@ class Go
...
 
       GPM_REGEXP = /^(.+)\s+(.+)$/
-      GOMOD_REGEX = /^(require\s+)?(.+)\s+(.+)$/
-      GOMOD_IGNORABLE_REGEX = /^(\/\/|module\s|go\s|exclude\s|replace\s|require\s+\(|\))/m
+      GOMOD_REPLACEMENT_SEPARATOR_REGEX = / => /
+      GOMOD_DEP_REGEX = /(?<name>\S+)\s?(?<requirement>[^ =>]+)?/
+      GOMOD_SINGLELINE_DEP_REGEX = /^(?<category>require|exclude|replace|retract)\s+#{GOMOD_DEP_REGEX}.*$/
+      GOMOD_MULTILINE_START_REGEX = /^(?<category>require|exclude|replace|retract)\s+\(/
+      GOMOD_MULTILINE_END_REGEX = /^\)/
+      GOMOD_MULTILINE_DEP_REGEX = /^#{GOMOD_DEP_REGEX}.*$/
```
tiegz, line 15: go.mod directives can be either singleline (`require foo v1.0.0`) or multiline (`require (\n foo v1.0.0\n)`) so this PR adds more intelligent parsing for both cases.

lib/bibliothecary/parsers/go.rb
```diff
@@ -7,8 +7,12 @@ class Go
       include Bibliothecary::Analyser
 
       GPM_REGEXP = /^(.+)\s+(.+)$/
-      GOMOD_REGEX = /^(require\s+)?(.+)\s+(.+)$/
-      GOMOD_IGNORABLE_REGEX = /^(\/\/|module\s|go\s|exclude\s|replace\s|require\s+\(|\))/m
+      GOMOD_REPLACEMENT_SEPARATOR_REGEX = / => /
```
johnbintz-tidelift, line 10: I was not sure what I was looking at initially. WDYT about using the `%r` syntax here to make it more clear this is a regular expression rather than...two divisions with an arrow operator?

lib/bibliothecary/parsers/go.rb
```diff
@@ -163,6 +194,34 @@ def self.map_dependencies(manifest, attr_name, dep_attr_name, version_attr_name,
           }
         end
       end
+
+      def self.go_mod_category_relative_dep(category, line, match)
```
johnbintz-tidelift, line 198: Can this use named parameters instead?

lib/bibliothecary/parsers/go.rb
```diff
@@ -120,18 +124,45 @@ def self.parse_dep_lockfile(file_contents, options: {}) # rubocop:disable Lint/U
...
-              requirement: match[3].strip || "*",
-              type: "runtime",
-            }
+          if line.match(GOMOD_MULTILINE_END_REGEX) # detect the end of a multiline
+            current_multiline_category = nil
+          elsif (match = line.match(GOMOD_MULTILINE_START_REGEX)) # or, detect the start of a multiline
+            current_multiline_category = match[1]
+          elsif (match = line.match(GOMOD_SINGLELINE_DEP_REGEX)) # or, detect a singleline dep
+            categorized_deps[match[:category]] << go_mod_category_relative_dep(match[:category], line, match)
+          elsif (current_multiline_category && match = line.gsub(/(\/\/(.*))/, "").match(GOMOD_MULTILINE_DEP_REGEX)) # otherwise, parse the multiline dep
```
johnbintz-tidelift, line 161: This `gsub` is for removing comments?

lib/bibliothecary/parsers/go.rb
tiegz: I figured it'd be clear since the var is suffixed with `_REGEX`, forward-slashes are the ubiquitous delimiter for regexes and it's listed above/below the other regexes, but I'm not opposed to using a different syntax here

lib/bibliothecary/parsers/go.rb
tiegz: yeah, I should probably extract that into a constant too

lib/bibliothecary/parsers/go.rb
tiegz: actually, I'm hesitant to start mixing syntaxes in bibliothecary, so I'll just replace the whitespaces with `\s` to make it more obvious

johnbintz-tidelift (approved):

PR #588, opened 2024-01-10
8.7.6:  go-resolved-dependencies.json: adds official support for "replace" directive 

tiegz:
(followup to https://github.com/librariesio/bibliothecary/pull/585)

as we learned today, `"replace"` directives are not reported by things like `go mod graph`, and are only reported by `go list` if you ask for the ".Replace" directive, since `"replace"` only replaces the source code of the replaced version and not the actual version reported as used by go.

this adds support for the "Replace" field in `go-resolved-dependencies.json` files such that we can replace the deps during Bibliothecary analysis of those files, as we did in #585 with `go.mod` files.

### DIFF
--- a/spec/fixtures/go-resolved-dependencies.json
+++ b/spec/fixtures/go-resolved-dependencies.json
@@ -3,104 +3,133 @@
       "Main": "true",
       "Version": "",
       "Indirect": "false",
-      "Scope": "runtime"
+      "Scope": "runtime",
+      "Replace": "<nil>"
     },
     {
       "Path": "cloud.google.com/go",
       "Main": "false",
       "Version": "v0.36.0",
       "Indirect": "true",
-      "Scope": "runtime"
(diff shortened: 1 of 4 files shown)


spec/fixtures/go-resolved-dependencies.json
```diff
@@ -3,104 +3,132 @@
...
-      "Scope": "runtime"
+      "Scope": "runtime",
+      "Replace": "<nil>"
     },
     {
       "Path": "github.com/stretchr/testify",
       "Main": "false",
       "Version": "v1.7.0",
       "Indirect": "false",
       "Scope": "test"
```
tiegz, line 118: NB: I've purposely left `"Replace"` out of this object to test that it's fine when it's missing.

wenottingham (approved):
idea seems good

mpace965 (approved):

PR #592, opened 2024-05-08
Parsers::NPM: handle dependencies on local paths in lockfiles

mpace965:
This PR changes `Parsers::NPM` so that it reports dependencies with [local path requirements](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#local-paths) in `package.json` manifests, `package-lock.json` (v2 or above), and `yarn.lock` lockfiles. It does so by adding a boolean-valued `local` key to the output.

It also de-dupes local path requirements from `package-lock.json` lockfiles, where they have two different entries.


lib/bibliothecary/parsers/npm.rb
```diff
@@ -149,6 +149,18 @@ def self.lockfile_preference_order(file_infos)
           ] + transform_tree_to_array(metadata.fetch("dependencies", {}))
         end.flatten(1)
       end
+
+      private_class_method def self.parse_name_from_package_lock_v2(name)
```
jhan217, line 153: 🎱 Any particular reason for the `_v2` naming?

jhan217 (approved):

lib/bibliothecary/parsers/npm.rb
mpace965: I'm working on a round of changes based on @tiegz's feedback above, and this method will go away actually. But the v2 naming here refers to the lockfile version

spec/fixtures/npm-lockfile-version-1/package-lock.json
mpace965, line 1: this is a fixture oops, should be testing `"lockfileVersion": 1`

tiegz (approved):

PR #593, opened 2024-05-24
Add support for pipdeptree JSON output in pip-dependency-graph.json

tiegz:
this adds support for a new dependency tree file generated by [pipdeptree](https://github.com/tox-dev/pipdeptree/blob/main/tests/render/test_text.py#L25) called "pip-dependency-graph.json". 

(pipdeptree calls these trees, but also warns that they may contain cycles, so we're calling them graphs)


spec/parsers/pypi_spec.rb
```diff
@@ -204,6 +204,35 @@
     })
   end
 
+  it "parses dependencies from pip-resolved-dependencies.txt, with tree" do
+    expect(described_class.analyse_contents("pip-resolved-dependencies.txt", load_fixture("pip_with_tree/pip-resolved-dependencies.txt"))).to eq({
+      platform: "pypi",
+      path: "pip-resolved-dependencies.txt",
+      dependencies: [
+        { name: "black", requirement: "==23.12.0", type: "runtime" },
+        { name: "aiohttp", requirement: "3.9.5", type: "runtime" },
```
tiegz, line 213: note how the regular requiremens.txt-compatible lines are parsed with "==" version range prefix, but the tree-syntax lines do not get that, because they're parsed from a string like `"├── aiohttp [required: >=3.7.4, installed: 3.9.5]"`. 

we could theoretically prefix them with "==" ourselves for consistency, but it doesn't seem like there's a big need since they both mean the same thing.

spec/parsers/pypi_spec.rb
```diff
@@ -204,6 +204,34 @@
     })
   end
 
+  it "parses dependencies from pip-dependency-tree.json" do
```
jhan217, line 207: 🎱 Should there be a couple of cases to test out the `.uniq` calls?

lib/bibliothecary/parsers/pypi.rb
```diff
@@ -252,18 +268,16 @@ def self.parse_requirements_txt(file_contents, options: {}) # rubocop:disable Li
...
-            match = line.delete(" ").match(REQUIREMENTS_REGEXP)
-            next unless match
-
+          elsif (match = line.delete(" ").match(REQUIREMENTS_REGEXP))
             deps << {
               name: match[1],
               requirement: match[-1] || "*",
               type: type,
             }
           end
```
jhan217, line 277: What things would fall into the `else` block? Should we test or alert on that condition?

lib/bibliothecary/parsers/pypi.rb
```diff
@@ -252,18 +268,16 @@ def self.parse_requirements_txt(file_contents, options: {}) # rubocop:disable Li
...
         end
-        deps
+
+        deps.uniq
```
jhan217, line 280: 🎱 Depending on how many dupes there are, could be more memory efficient to initialize deps as a Set object and `.to_a` the return here

lib/bibliothecary/parsers/pypi.rb
tiegz: this change should be functionally the same (the `else` + `next` is basically just `elsif` here). I just decided to combine the 3 lines into 1 line randomly.

lib/bibliothecary/parsers/pypi.rb
tiegz: I thinkkk `pipdeptree`'s top-level of deps in the JSON should never have duplicates, but I just included it here to be extra safe 🤷🏻

spec/parsers/pypi_spec.rb
tiegz: this comment is irrelevant now that I'm using the JSON instead of requirements-compatible file

spec/parsers/pypi_spec.rb
tiegz: done-so

jhan217 (approved):

PR #596, opened 2024-07-03
[10.0.0] Return a new Dependency class instead of raw Hash objects, from analyses' :dependencies result.

tiegz:
This defines a new `Bibliothecary::Dependency` class, and proposes to return instances of that object from parsed manifests, instead of Hashes.

### Current behavior (<10.0)

Bibliothecary returns a list of dependency Hashes in its analyses, e.g.:

```
analysis = Bibliothecary.analyse_file("package.json", "package.json")
analysis[:dependencies]
=> [{ name: "jquery", requirement: ">= 1.9.1", type: "runtime" }]
```

### New behavior (>= 10.0)

```
analysis = Bibliothecary.analyse_file("package.json", "package.json")
analysis[:dependencies]
=> [Bibliothecary::Dependency.new(name: "jquery", requirement: ">= 1.9.1", type: "runtime")],
```

### Benefits

Having a dedicated `Dependency` class will:
* ... define a contract between Bibliothecary and its downstream users
* ... raise errors when a field is accessed on a dependency that doesn't exist
* ... documents the possible fields of dependencies
* ... sets requirements for dependencies (`"name"` and `"requirement"`)

### Downsides

* breaking change (this would be merged into the `10-0` branch)
* more verbose Bibliothecary code
* doesn't benefit from autocomplete for most people, since most IDEs aren't aware of required gems

### TODO

* [x] swift_pm.rb
* [x] shard.rb
* [x] rubygems.rb
* [x] pypi.rb
* [x] pub.rb
* [x] packagist.rb
* [x] nuget.rb
* [x] npm.rb
* [x] meteor.rb
* [x] maven.rb
* [x] julia.rb
* [x] hex.rb
* [x] haxelib.rb
* [x] hackage.rb
* [x] go.rb
* [x] elm.rb
* [x] dub.rb
* [x] cran.rb
* [x] cpan.rb
* [x] conda.rb
* [x] cocoapods.rb
* [x] clojars.rb
* [x] carthage.rb
* [x] cargo.rb
* [x] bower.rb
* [x] CycloneDX and SPDX multi parsers
* [x] CHANGELOG.md file (or README section) that describes how to upgrade apps for this change in 10.0


lib/bibliothecary/dependency.rb
```diff
@@ -0,0 +1,39 @@
+module Bibliothecary
+  # @attr [String] name The name of the package, e.g. "ansi-string-colors"
```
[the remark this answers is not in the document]
tiegz: yard has you put these in the class docstring, although I should probably rename these to `@attr_reader` now that you point it out: https://rubydoc.info/gems/yard/file/docs/Tags.md#attr_reader

lib/bibliothecary/dependency.rb
```diff
@@ -0,0 +1,39 @@
+module Bibliothecary
+  # @attr_reader [String] name The name of the package, e.g. "ansi-string-colors"
+  # @attr_reader [String] requirement The version requirement of the release, e.g. "1.0.0" or "^1.0.0"
+  # @attr_reader [String] type The type of dependency, e.g. "runtime" or "test"
+  # @attr_reader [Boolean] direct Is this dependency a direct dependency (vs transitive dependency)? (default: nil)
+  # @attr_reader [Boolean] deprecated Is this dependency deprecated? (default: nil)
```
tiegz, line 6: Q: opinions on defaulting these to something, instead of `nil`?

lib/bibliothecary/parsers/packagist.rb
```diff
@@ -25,22 +25,29 @@ def self.mapping
...
+          Dependency.new(
             name: dependency["name"],
-            requirement: dependency["version"],
+            requirement: requirement,
             type: "development",
-          }.tap do |result|
-            # Store Drupal version if Drupal, but include the original manifest version for reference
-            result[:original_requirement], result[:requirement] = result[:requirement], dependency.dig("source", "reference") if is_drupal_module(dependency)
+            original_requirement: original_requirement
+          ).tap do |result|
```
macowie, line 50: Orphan `tap` block here?

lib/bibliothecary/dependency.rb
macowie: Were you thinking some explicit `unknown` value, or making an assumption (eg default to direct = true)? For `type` and `direct` I'm not sure there's an obvious default to me. `deprecated = false` could make sense though.

Introducing that kind of change makes this less of drop-in replacement, but I suppose this could be an opportune time to return something with less ambiguity?

tiegz: good points, we could always return later to set defaults. I'll just keep the `deprecated` and `optional` defaults of `false` in here since they seem less contentious, and we could think about the others later.

tiegz: actually wait, those are both fields specific to certain platforms (Golang, Maven, etc), so maybe I should keep them `nil` defaults for now.

macowie (approved):

PR #603, opened 2024-09-24
feat: [sc-45241] Import json formatted SPDX sboms

tradiff:
Story details: https://app.shortcut.com/tidelift/story/45241

- Allow *.spdx.json parsing
- Change *.spdx parser to prefer purl for the name and version, only using the spdx fields if needed
- Some cleanup


.tidelift.yml
```diff
@@ -1,2 +0,0 @@
-ci:
-  extra_ignore_directories: [ spec/fixtures ]
```
katzj, line 2: I think we want to keep this to not analyze the subdirectories

lib/bibliothecary/multi_parsers/spdx_json.rb
```diff
@@ -0,0 +1,80 @@
...
+
+        manifest["packages"]&.each do |package|
+          package_name = package["name"]
+          package_version = package["versionInfo"]
+
+          platform = nil
+          package["externalRefs"]&.each do |ref|
+            if ref["referenceType"] == "purl"
+              purl = ref["referenceLocator"]
+              platform ||= get_platform(purl)
```
katzj, line 62: I feel like if we have a purl, we might get a better package name from it than we do from the name field. At least that seemed to be the case when I briefly looked a week or two ago

katzj: Yeah, we definitely want to use the name from the purl if there is one. Looking at an example spdx json sbom generated from github has 
```
    {
      "SPDXID": "SPDXRef-nuget-AWSCRT-AUTH-0.4.4",
      "name": "nuget:AWSCRT-AUTH",
      "versionInfo": "0.4.4",
      "downloadLocation": "NOASSERTION",
      "filesAnalyzed": false,
      "licenseConcluded": "Apache-2.0",
      "supplier": "NOASSERTION",
      "externalRefs": [
        {
          "referenceCategory": "PACKAGE-MANAGER",
          "referenceLocator": "pkg:nuget/AWSCRT-AUTH@0.4.4",
          "referenceType": "purl"
        }
      ]
    },
```

The name has the platform embedded in it but there's no real structure to what the name is supposed to be that I can see. So I think if there's a purl, we want to use the information out of that and then fallback to the non-purl fields

tradiff: Sounds like that should also apply to the tag:value spdx parser. I'll update both.

katzj (approved):

PR #607, opened 2024-11-06
Fix using parent props for deps

katzj:
Refactoring of this was returning too early and not using any parent properties that are specified in the pom file.

property_value now returns nil for cases where it can't find the value and extract_property is properly using the fallback value in that case rather than requiring an interpolation of what it should have been

### DIFF
--- a/lib/bibliothecary/parsers/maven.rb
+++ b/lib/bibliothecary/parsers/maven.rb
@@ -425,7 +425,6 @@ module Bibliothecary
         # the xml root is <project> so lookup the non property name in the xml
         # this converts ${project/group.id} -> ${group/id}
         non_prop_name = property_name.gsub(".", "/").gsub("project/", "")
-        return "${#{property_name}}" if !xml.respond_to?("properties") && parent_properties.empty? && xml.locate(non_prop_name).empty?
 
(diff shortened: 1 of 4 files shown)


lib/bibliothecary/parsers/maven.rb
```diff
@@ -425,7 +425,6 @@ def self.property_value(xml, property_name, parent_properties)
         # the xml root is <project> so lookup the non property name in the xml
         # this converts ${project/group.id} -> ${group/id}
         non_prop_name = property_name.gsub(".", "/").gsub("project/", "")
-        return "${#{property_name}}" if !xml.respond_to?("properties") && parent_properties.empty? && xml.locate(non_prop_name).empty?
```
tiegz, line 428: to confirm: before, we were returning "${some_property}" if there were no props and no parent props, but now we'll return nil?

if that's correct, the downside seems like it'll be harder to debug uninterpolated props?

katzj: Right. But extract_property (which is the only thing that calls property_value) has a thing to handle property_value being a nil return (falling back to the value).

This is how the code originally worked but in some refactoring along the way, it got confused

tiegz (approved):

mikeyoung85 (approved):

PR #613, opened 2025-01-30
12.0.1: Enable default Rubocop rules

tiegz:
Making spec changes in https://github.com/librariesio/bibliothecary/pull/612 was pretty arduous because we didn't have auto-formatting rules for spacing/indentation/etc.

These are decisions better made by the computer, so this enables all the default Rubocop rules. Most of the violations were auto-correctable, and I did a few manually, so it wasn't too bad.

Also, bump the minor version to 12.1.0


Gemfile
```diff
@@ -1,9 +1,20 @@
...
 gemspec
 
+group :development do
+  gem "pry"
+  gem "rake", "~> 12.0"
+  gem "rspec", "~> 3.0"
+  gem "rubocop", "~> 1.71"
+  gem "rubocop-rails"
+  gem "webmock"
+end
```
tiegz, line 15: TIL that we should use `add_dependency` in the gemspec, avoid `add_runtime_dependency`/`add_development_dependency`, and just put dev dependencies n Gemfile.

tradiff (approved):

PR #621, opened 2025-03-14
Add Support for bun.lock File

mikeyoung85:
[Bun](https://bun.sh/) is another Javascript package management tool and has recently [changed their lockfile](https://bun.sh/blog/bun-lock-text-lockfile) format to a JSON based text file. This is the default as of version 1.2 released earlier this year. This PR adds support to read from this file to find NPM dependencies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add support for bun.lock files
+
 ### Changed
 
 ### Removed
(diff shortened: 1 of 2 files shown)


bibliothecary.gemspec
```diff
@@ -23,6 +23,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency "commander"
   spec.add_dependency "deb_control"
+  spec.add_dependency "json", "~> 2.8"
```
mikeyoung85, line 26: The ruby json implementation added the ability to accept json with trailing commas in [version 2.8.0](https://github.com/ruby/json/releases/tag/v2.8.0). The `bun.lock` file uses trailing commas in their output and in order to read their file correctly we need this updated version of the json gem that is usually bundled with ruby.

tiegz (approved):

tiegz:
I held off on bumping the version here, because I have a small PNPM fix to roll out today too

PR #624, opened 2025-04-29
12.1.6: Don't raise an error if PNPM lockfile has no devDependencies section.

tiegz:
It's possible that a `pnpm-lock.yaml` may not have a `devDependencies` section, so fall back to `{}` in those cases instead of raising an error. For example:

```
lockfileVersion: '9.0'

importers:

  .:
    dependencies:
      ansicolor:
        specifier: ^2.0.3
        version: 2.0.3

...
```


wenottingham (approved):
sure!

PR #629, opened 2025-05-23
12.1.10: Pypi: normalize package names in Poetry manifests, and store the original too.

tiegz:
Poetry will take a package name from `pyproject.toml`, and store the [normalized name according to PyPa's rules](https://packaging.python.org/en/latest/specifications/name-normalization/#name-normalization) in `poetry.lock` without providing the original. This change normalizes all names for Poetry manifests so they're consistent, and stores the original in `Dependency#original_name` if it differs.

### DIFF
--- a/lib/bibliothecary/parsers/pypi.rb
+++ b/lib/bibliothecary/parsers/pypi.rb
@@ -101,21 +101,21 @@ module Bibliothecary
 
         # Parse poetry [tool.poetry] deps
         poetry_manifest = file_contents.fetch("tool", {}).fetch("poetry", {})
-        deps += map_dependencies(poetry_manifest["dependencies"], "runtime", options.fetch(:filename, nil))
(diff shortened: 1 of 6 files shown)


lib/bibliothecary/parsers/pypi.rb
```diff
@@ -144,17 +144,25 @@ def self.map_dependencies(packages, type, source = nil)
...
                 local: local
               )
             end
           else
+            # Poetry normalizes names in lockfiles but doesn't provide the original, so we need to keep
+            # track of the original name so the dep is connected between manifest+lockfile.
+            normalized_name = normalize_name(name)
             Dependency.new(
-              name: name,
+              name: normalize_name ? normalized_name : name,
```
katzj, line 164: Shouldn't this be
```suggestion
              name: normalize_named ? normalized_name : name,
```

tiegz: (the `normalized_name` computation on line 162 is just a way to DRY it up so we don't have to call `normalize_name()` twice too)

katzj (approved):

PR #634, opened 2025-06-23
[13.1.0] Convert parsed Nuget requirements to SemVer-style requirements

tiegz:
For the given `*.csproj` requirement, return `">= 1.2.3"` instead of `"1.2.3"`:

`<PackageReference Include="SomePackage" Version="1.2.3" />`

This change also applies all the other range equivalencies from the [Nuget docs](https://learn.microsoft.com/en-us/nuget/concepts/package-versioning?tabs=semver20sort#version-ranges) so that we return [NPM-style](https://docs.npmjs.com/cli/v6/using-npm/semver) SemVer requirements instead of Nuget-style requirements.

This notably makes the Nuget requirements compatible with the [semantic_range](https://github.com/librariesio/semantic_range) gem too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
+## [13.1.0] - 2025-06-23
+
+### Changed
+
+- Fix any parsed nuget requirements by translating them from Nuget requirements to unofficial SemVer requirements (i.e. basically NPM-style ranges).
+
 ## [13.0.0] - 2025-06-17
 
 ### Removed
--- a/lib/bibliothecary/parsers/nuget.rb
+++ b/lib/bibliothecary/parsers/nuget.rb
@@ -2,6 +2,7 @@
 
 require "ox"
 require "json"
+require "bibliothecary/parsers/nuget/nuget_requirement"
 
 module Bibliothecary
   module Parsers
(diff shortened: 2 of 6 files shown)


lib/bibliothecary/parsers/nuget/nuget_range.rb
```diff
@@ -0,0 +1,64 @@
...
+        # @return [String] The converted SemVer requirement string
+        # @raise [ArgumentError] If the requirement format is invalid
+        def self.convert_to_semver_requirement(requirement)
+          case requirement
+          when MINOR_WILDCARD_REGEXP
+            "~> #{::Regexp.last_match(1)}.0"
+          when PATCH_WILDCARD_REGEXP
+            ">= #{::Regexp.last_match(1)}.0"
+          when FULL_WILDCARD_REGEXP
+            ">= 0"
```
tiegz, line 36: I'm on the fence about this: we've solidified around a default "*" wildcard for most Bibliothecary parsers, which is effectively the same as ">= 0".

So returning ">= 0" is more semantically consistent, but "*" is also safe and would be more consistent with Bibliothecary.

Thoughts?

wenottingham:
> For the given `*.csproj` requirement, return `">= 1.2.3"` instead of `"1.2.3"`:
> 
> `<PackageReference Include="SomePackage" Version="1.2.3" />`
> 
> This change also applies all the other range equivalencies from the [Nuget docs](https://learn.microsoft.com/en-us/nuget/concepts/package-versioning?tabs=semver20sort#version-ranges) so that we return [NPM-style](https://docs.npmjs.com/cli/v6/using-npm/semver) SemVer requirements instead of Nuget-style requirements.
> 
> This notably makes the Nuget requirements compatible with the [semantic_range](https://github.com/librariesio/semantic_range) gem too.

Does this make our results in the absence of a lockfile worse?

tiegz:
> Does this make our results in the absence of a lockfile worse?

Yes, but it'll make them more accurate too

tiegz:
I might close this one after some reflection:
* we don't have an official syntax for the "requirement" field, so maybe Nuget syntax is fine
*

PR #635, opened 2025-07-03
13.0.1: Handle Windows-style newlines in v1 yarn.lock files

tiegz:


lib/bibliothecary/parsers/maven.rb
```diff
@@ -274,7 +274,7 @@ def self.parse_maven_resolved(file_contents, options: {})
       def self.parse_maven_tree_items_with_depths(file_contents)
         file_contents
           .gsub(ANSI_MATCHER, "")
-          .gsub(/\r\n?/, "\n")
+          .encode(universal_newline: true)
```
tiegz, line 277: Switched Maven to use this too, docs here: https://github.com/ruby/ruby/blob/master/doc/encodings.rdoc#encoding-options-

PR #636, opened 2025-07-14
13.1.0: Return Maven projects as Dependencies themselves, from "maven-dependency-tree.dot" lockfiles.

tiegz:
This enables bibliothecary to denote what the project name/version for a list of dependencies is, by:

* adding `Dependendency#project` boolean, which is true when the dependency is the project itself (default: nil)
* Maven: start returning a Dependency for the project itself from `maven-dependency-tree.dot` manifests, which will be marked `project: true`

Upsides: 
* it's much simpler than the alternative of returning a new "project_name" field for each analysis result, which would require changing every manifest-parsing method to support. 
* this might also be useful for Gradle projects, etc

Downsides:
* it's debatable if the project itself can be considered a dependency.

Thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
+## [13.1.0] - 2025-07-15
+
+### Added
+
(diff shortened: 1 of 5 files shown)


tiegz:
After giving this some time, I'm wondering if "returning the project name/version as part of analysis results" is actually a better fix.

PR #639, opened 2025-07-22
14.0.0: introduce a ParserResult class to wrap parser results

tiegz:
(alternative to https://github.com/librariesio/bibliothecary/pull/636 )

* adds `Bibliothecary::ParserResult`, which contains `dependencies` array, and a new `project_name` field as metadata about the manifests being parsed (which default to `nil`)
* changes the return type of  `Bibliothecary::Parser` parser methods:
  * before: returned `Array<Bibliothecary::Dependency>`
  * after: will return `DependenciesResult`
* return `project_name` for the following files: 
  * `maven-dependency-tree.txt` 
  * `maven-dependency-tree.dot`
  * (we can add more later)


lib/bibliothecary/dependencies_result.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+module Bibliothecary
+  # DependenciesResult bundles together a list of dependencies and the project name.
+  #
+  # @attr_reader [Array<Dependency>] dependencies The list of Dependency objects
+  # @attr_reader [String,nil] project_name The name of the project
+  class DependenciesResult
```
macowie, line 8: Yeah, ParserResult makes it clearer where its coming from and _why_ it contains what it does.

macowie (approved):

PR #640, opened 2025-07-23
Require the "platform" field to be set on Bibliothecary::Dependency.

tiegz:
This is a breaking change and should make it into [14.0.0](https://github.com/librariesio/bibliothecary/pull/639):

* require `Bibliothecary::Dependency#platform` to be present
* update all the parsers to pass `platform: platform_name` when instantiating `Bibliothecary::Dependency`

Reasoning: this makes it easier to pass around Dependency objects downstream without having to group them by platform, set their platform manually, etc.

### DIFF
--- a/lib/bibliothecary/parsers/cocoapods.rb
+++ b/lib/bibliothecary/parsers/cocoapods.rb
@@ -54,7 +54,7 @@ module Bibliothecary
 
       def self.parse_podspec(file_contents, options: {})
         manifest = Gemnasium::Parser.send(:podspec, file_contents)
-        parse_ruby_manifest(manifest, "cocoapods", options.fetch(:filename, nil))
+        parse_ruby_manifest(manifest, platform_name, options.fetch(:filename, nil))
       end
 
       def self.parse_podfile(file_contents, options: {})


tiegz:
> Does this have any significant impact on memory usage? But makes sense to me

It shouldn't affect the Big-O because it just adds another string  we're allocating N times (N=number of deps) so it goes from best-case of 2N strings (name+requirement) to 3N strings (name+requirement+platform), which is just O(N).

But it does increase memory usage of strings overall, so I just pushed an improvement that freezes the "platform_name" string so that we don't re-allocate the same string over and over:

### Before

``` ruby
> Bibliothecary::Parsers::Maven.platform_name.object_id
197720
> Bibliothecary::Parsers::Maven.platform_name.object_id
198140
> Bibliothecary::Dependency.new(name: "foo", requirement: "1.0.0", platform: Bibliothecary::Parsers::Maven.platform_name).platform.object_id
199760
> Bibliothecary::Dependency.new(name: "foo", requirement: "1.0.0", platform: Bibliothecary::Parsers::Maven.platform_name).platform.object_id
201180
```

### After

``` ruby
> Bibliothecary::Parsers::Maven.platform_name.object_id
200920
> Bibliothecary::Parsers::Maven.platform_name.object_id
200920
> Bibliothecary::Dependency.new(name: "foo", requirement: "1.0.0", platform: Bibliothecary::Parsers::Maven.platform_name).platform.object_id
200920
> Bibliothecary::Dependency.new(name: "foo", requirement: "1.0.0", platform: Bibliothecary::Parsers::Maven.platform_name).platform.object_id
200920
```

Thx for prompting me to look at that!

macowie (approved):

PR #645, opened 2025-09-30
14.0.3: For pom.xml files, fall back to "type: nil" instead of "type: 'runtime'" if scope is missing

tiegz:
When parsing a Maven pom.xml and lockfile pair, the lockfile will hold the source of truth for `<scope>`. The manifest may not have that value, and the parent pom may not be available to resolve `<scope>`. Bibliothecary should just return `nil` instead of defaulting to "runtime" and let the client use the lockfile to find the actual value to avoid erroneous `<scope>` assumptions.

(also, a few Rubocop lints)


tiegz:
@katzj true, this could also just be seen as a bugfix TBH. I'll update the version

PR #646, opened 2025-10-02
Nuget: ignore <reference> dependencies if they don't have a version.

tiegz:
In .csproj files for Nuget, we can be pickier by excluding some `"<Reference>"` tags since they're mostly for referencing assemblies. If there's a version, that means it'll get downloaded by Nuget (and we want it), and if not then it's a built-in or on-disk assembly.


wenottingham (approved):
ruby syntax confuses me, but test case implies right thing looking at file

PR #647, opened 2025-10-02
[14.2.0] Stop assuming "runtime" scope for deps from yarn.lock

tiegz:
This is consistent with our [recent pom.xml change](https://github.com/librariesio/bibliothecary/commit/a83d35f4cb593bb576683e4057322e1f503ab577) of returning `type: nil` when we can't be certain about a scope. Yarn.lock files don't give you info about the scope, so we should return nil and let the downstream clients do the math.


katzj (approved):
Obviously we'll need to update depci when pulling in the new version

PR #666, opened 2026-02-20
[15.3.1] Prefix subproject names in gradle-dependencies-q.txt with "subproject"

tiegz:
e.g. for project names, `subproject:app` will be the name that we return for this sample line:

`Project ':app'`

e.g. for dependencies, `subproject:utilities` will be the name that we return for this sample line:

`+--- project :utilities`

This makes the names for these dependencies more Maven-friendly without having to manipulate them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
+## [15.3.1]
+
+### Changed
+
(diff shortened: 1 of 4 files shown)


katzj:
Does this end up being confusing, similar to how `internal:` was previously?

tiegz:
@katzj yeah, the goal is to not show these at all, but in some cases they might still show up so "subproject" seemed like an intuitive name

katzj (approved):

PR #667, opened 2026-02-25
15.5.0: move gradle-dependencies-q.txt line parsing into a single regexp.

tiegz:
The goal here was to reduce String allocations as much as possible, and it worked. The code is simpler too.


spec/parsers/maven_spec.rb
```diff
@@ -752,6 +752,38 @@
                ))
     end
 
+    it "parses version range to resolved version syntax" do
+      version_range = "+--- org.codehaus.woodstox:stax2-api:[3.0.4, 3.5.0) -> 3.1.4"
```
tiegz, line 756: version ranges was something we hadn't picked up before

PR #669, opened 2026-03-12
[15.5.0] Npm: adds yarn_workspace_dependencies option that includes workspace deps

tiegz:
When enabled, the parser will return workspace dependencies from `yarn.lock` files:

`Bibliothecary::Parsers::Npm.analyse_contents("yarn.lock", load_fixture("yarn-workspace/yarn.lock"), options: { yarn_workspace_dependencies: true }`


lib/bibliothecary/parsers/npm.rb
```diff
@@ -231,20 +226,26 @@ def self.parse_v1_yarn_lock(contents, source = nil)
           end
       end
 
-      def self.parse_v2_yarn_lock(contents, source = nil)
+      # v2+ yarn lockfiles have a __metadata section and are YAML-compatible
+      def self.parse_v2_yarn_lock(contents, source = nil, yarn_workspace_dependencies: false)
         parsed = YAML.load(contents)
         parsed = parsed.except("__metadata")
+
+        if !yarn_workspace_dependencies && parsed.any? { |packages, info| info["version"].to_s.include?("use.local") && packages.include?("workspace") }
```
mikeyoung85: Could this be more specific and check the version is exactly `0.0.0-use.local`? I'm not sure if that ever changes or if yarn will always use that string for the version.

The name could also be a little more specific and use `"@workspace:"` to make sure it is the yarn generated name, but I would imagine the version check before would catch most things correctly anyways.

tiegz: Yeah, that's probably a good idea: the source code seems to show that this is always true: https://github.com/yarnpkg/berry/blob/7ccf6e3da18d77d36a426d004babf8553639defa/packages/yarnpkg-core/sources/scriptUtils.ts#L156

PR #670, opened 2026-04-28
Add Lean parser for Lake package files

andrew:
Adds support for Lean 4 packages managed by Lake. Three file types under the `lean` platform:

- `lakefile.toml` (manifest): parses `[[require]]` tables. Handles `scope`, `version`, `rev`, `git`, `path`, and inline `source` tables. Path-only deps are skipped.
- `lakefile.lean` (manifest): regex over the `require` DSL. Covers Reservoir form (`require "scope" / "name" @ git "rev"`), direct git form (`require Name from git "url" @ "rev"` including multiline), and guillemet names (`«Name»`). Strips `--` line comments and skips local path deps.
- `lake-manifest.json` (lockfile): parses `packages[]`, using the resolved `rev` SHA as requirement and `!inherited` as the direct flag. Path-type entries are skipped.

Dependency names use `scope/name` when a Reservoir scope is present (e.g. `leanprover-community/batteries`) and the bare package name for direct git deps without a scope. This matches the naming on Reservoir (reservoir.lean-lang.org).

Fixtures based on real `aesop`, `doc-gen4`, and `mathlib4` files.


andrew:
Wrong repo, sorry — meant for ecosyste-ms/bibliothecary.

